PR #1, opened 2018-08-13
Import from main documentation branch

Nazgolze:
This is imported from the docs staging site.

Just to make sure that we don't lose any content when we destroy docs.liskdev.net.

I don't intend for this to get merged as is (or maybe not at all).

At first glance, this looks bigger than it is, looks like some of this restructuring changes that, if we're going to that releases and have different versions that way, don't necessarily need to happen (v1.0 vs v1.1).


Tschakki (commented):
The following changes are good:
- 3-lisk-core/3-user-guide/2-administration/page-documentation-content.txt
- 3-lisk-core/3-user-guide/3-api/1-1-0/page-documentation-content.txt
- 3-lisk-core/3-user-guide/3-api/2-0-9-vs-1-0/page-documentation-content.txt
- 4-lisk-elements/3-user-guide/page-documentation-content.txt
- 5-lisk-commander/3-user-guide/2-commands/page-documentation-content.txt

The rest does not need to be merged because we decided to only keep the latest documentation on the webpage, as @Nazgolze stated correctly.

karmacoma:
Closing in favour of #2

PR #158, opened 2019-01-21
Add section about versioning schemes - Closes #62

### LINKED ISSUE #62: Mention something about Lisk core using SemVer
It doesn't appear that the docs discuss the Lisk core versioning approach so some users might get confused about version compatibility between nodes. See https://github.com/LiskHQ/lisk/issues/753

We could add a paragraph somewhere to mention that Lisk core follows SemVer conventions; maybe we should also link to an external resource so that the user can read about it in more details.

Maybe this also applies to other projects as well.

Tschakki:
Closes #62

### DIFF
--- a/introduction.md
+++ b/introduction.md
@@ -3,8 +3,8 @@
 ## Table of contents
 
 - [Introduction](#lisk-core-documentation)
-  - [Lisk Core Version Reference](#versions)
-  - [Versioning Scheme](#versioning-scheme)
+  - [Versions](#versions)
+  - [Versioning Schemes](#versioning-schemes)
   - [What is Lisk Core?](#what-is-lisk-core)
     - [Networks](#networks)
     - [Components](#components)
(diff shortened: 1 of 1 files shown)


introduction.md
```diff
@@ -4,6 +4,7 @@
 
 - [Introduction](#lisk-core-documentation)
   - [Lisk Core Version Reference](#versions)
+  - [Versioning Scheme](#versioning-scheme)
```
AndreasKendziorra: #versioning-scheme -> #versioning-schemes

introduction.md
```diff
@@ -37,6 +38,37 @@ Version | Release date <br> (yy/mm/dd)| Documentation reference
...
+```
+v<MAJOR>.<MINOR>.<PATCH> 
+
+v     - Version prefix
+MAJOR - Breaking change
+MINOR - New feature
+PATCH - Bug fix
+```
+
+The *Software Implementation Versioning* follows the popular SemVer scheme and gives a quick overview for developers about breaking and non-breaking changes in the software.
```
AndreasKendziorra: We should use consistently "Semver" or "SemVer"

introduction.md
```diff
@@ -37,6 +38,37 @@ Version | Release date <br> (yy/mm/dd)| Documentation reference
...
+
+### Protocol Versioning
+
+The protocol version is denoted by two digits, `H.S.`.
+The first digit, `H`, depends on the number of hard forks and is incremented with every hard fork.
+`S` represents the number of soft forks since the last hard fork.
+
+> The initial protocol version 1.0 is defined to be the one that was implemented by Lisk Core v1.0.0.
+
+The *Protocol Versioning* theme is used e.g. in P2P Communication between Lisk Core nodes, in order to determine, if the nodes have compatible versions of the Lisk protocol implemented.
```
AndreasKendziorra: Maybe "The protocol version is used" instead of "The *Protocol Versioning* theme is used"?

AndreasKendziorra (changes requested):
Besides the mentioned suggestions, it might be worth to add a "protocol version" column in the table in section "Versions". Then, one knows easily which protocol version is implemented by which software.

2snEM6 (approved):
No problems for me. Very clear :)

AndreasKendziorra (approved):
Looks good to me. But @MaciejBaj should confirm that v.1.5.0 does not imply a soft fork.

PR #267, opened 2019-06-18
adding getTimestamp function to cashback tutorial, Closes #266

### LINKED ISSUE #266: Add time getTimestamp function to print cashback script
Adding a function to generate a timestamp for the cashback transaction to prevent the transaction being rejected if called multiple times.

RachBLondon:
if this is OK, ill also make a PR on the code examples 👍 

Closes #266


Tschakki (approved):
We need to include this in the example apps as well imho :)

PR #268, opened 2019-06-19
Use correct link to cashback example

tobiaslins:


Tschakki:
This is already fixed in current `development` branch.

PR #294, opened 2019-07-03
Remove broadcastLimit from chain and add emitPeerLimit in network module

ishantiw:
### Description

`modules.chain.broadcasts.broadcastLimit` (previously used only for blocks broadcasts) was changed to `modules.network.emitPeerLimit` and behavior is now different.

We remove `broadcastLimit` from chain and add `emitPeerLimit` in network in the config

Closes https://github.com/LiskHQ/lisk-sdk/issues/3905


Tschakki (approved):
Thanks! :)

PR #306, opened 2019-07-16
Add temppath parameter

michielmulders:
The `tempPath` parameter is used for defining the location for storing the temporary pid and socket files.
This property has been introduced because of the restricted length for unix domain sockets (104 on MacOs and 108 on Linux) so users can change the location to a shorter path if needed. (Introduced in 2.1)

Related PR: https://github.com/LiskHQ/lisk-sdk/pull/3957


Tschakki (approved):
thank you!

PR #374, opened 2019-12-01
Update link to PostgreSQL documentation

johnalexhebert:
Lisk Core uses version 10 of psql. Updated link should point to https://www.postgresql.org/docs/10/static/app-psql.html .


Tschakki (approved):
Thank you!

PR #382, opened 2019-12-23
Add a git checkout master step to the Docker docs

nerdvibe:
If you follow the docker docs, it will automatically checkout on `development` branch, which is not the latest stable version of Lisk Core and it will not coldstart.

I'd suggest to checkout to `master` branch in order to make sure that the user is on the latests stable version.


Tschakki (approved):
Thank you!

PR #397, opened 2020-01-17
Add API spec to SDK docs, closes #394

### LINKED ISSUE #394: API spec for SDK
**Describe the solution you'd like**
Only the Lisk Core API is described via swagger-ui at the moment. There should be a seperate API spec for the Lisk in Asciidoc.

Tschakki:
Closes #394


chris529 (approved):
Looks good, it is definitely understandable although there are numerous grammar/punctuation errors, etc, which can be fixed at a later date.

PR #568, opened 2020-04-23
Insert missing image and update case changes in Core docs - closes #567 

### LINKED ISSUE #567: Insert missing image and update case changes in Core docs
Lisk Core product image is missing / Node, Peer should be small letters. Important at the top should be click here not Click here.

Refer to parent issue #533

chris529:
closes #567


Tschakki (approved):
LGTM

PR #594, opened 2020-05-07
Insert missing image and update case changes in Core docs 2.1.6 - closes #592

### LINKED ISSUE #592: Insert missing image and update case changes in Core docs 2.1.5
Lisk Core product image is missing / Node, Peer should be small letters. Important at the top should be click here not Click here..
Parent issue #591

chris529:
closes #592


Tschakki (approved):
Thanks!

PR #646, opened 2020-06-24
Update config docs - Closes #641

### LINKED ISSUE #641: Update config page, create config reference
## Which content is outdated?

- Needs to be updated before the 1.0 release
- (Configuration) we should focus on Lisk node and logging and move the rest to the config reference page.
- (Configuration) LISK_CORE_WS is the same as LISK_CORE_HTTP, only protocol is different. If not given, then LISK_CORE_HTTP is taken. That could go to config reference page.
- (Configuration) GEOIP_JSON example is wrong. The end-user has to pay subscription at a 3rd party provider such as [Maxmind](https://www.maxmind.com/en/geoip-demo) or find an equivalent service.
- (Configuration) LISK_STATIC can be replaced by the user to have a custom list of account aliases. Falls under config reference page.

Tschakki:
Closes #641


modules/ROOT/pages/references/configuration.adoc
```diff
@@ -0,0 +1,111 @@
+= Configuration
+Mona Bärenfänger <mona@lightcurve.io>
+:description: A general reference of the Lisk Service configuration.
+
+:url_config: configuration.adoc
+
+The Lisk Service configuration reference comprises a complete list of all available configuration options, including examples and descriptions.
```
chris529, line 7:
```suggestion
The Lisk Service configuration reference comprises of a complete list of all available configuration options, including examples and descriptions.
```

modules/ROOT/pages/references/configuration.adoc
```diff
@@ -0,0 +1,111 @@
...
+
+Check out the xref:{url_config}[Lisk Service config guide] for more information, how to configure Lisk Service.
```
chris529:
```suggestion
Check out the xref:{url_config}[Lisk Service config guide] for more information on how to configure Lisk Service.
```

chris529 (commented):
Looks very good, just a couple of very minor points.. :-)

modules/ROOT/pages/references/configuration.adoc
Tschakki: Hm, I think we had the discussion about adding an `of` here already in the past, and I thought we agreed to not add it. But to be sure, I now researched the topic and concerning to this dictionary it is ungrammatical to add the "of" here: http://www.chambersharrap.co.uk/chambers/features/chref/chref.py/main?query=comprise&title=21st
Also the version of `comprise` seems much more common than `comprise of` (in my ears for example `comprise of` sounds a little weird, too).

chris529: Hi Mona, Yes you are correct it is ironic really, as to me personally it sounds better using "of" in these cases. However you are correct, as it is also right without the "of' OK, lets leave it as you had it originally :-)

chris529 (approved):

PR #735, opened 2020-09-17
Align docker container name - Closes #727

### LINKED ISSUE #727: Docker container names are not the same from PostgreSQL from Setup to Tutorials
## Which content is outdated?

In [this link](https://lisk.io/documentation/lisk-sdk/setup.html#postgresql) , we say that to install PostgreSQL with docker we can just run the following command:

```bash
docker run --name lisk_db -p 5432:5432 -e POSTGRES_USER=lisk -e POSTGRES_PASSWORD=password -e POSTGRES_DB=lisk_dev -d postgres:10
```

But then [here](https://lisk.io/documentation/lisk-sdk/tutorials/hello-world.html#_dependencies) in the docker part, the command is not the same:

```bash
docker exec -ti lisk_sdk_db psql -h localhost -U lisk -d postgres -c "DROP DATABASE lisk_dev"
```

For a better experience, its better if you have the same name, otherwise users might get confused why it is not working. Since Docker will throw an error mentioning that there is no container names lisk_sdk_db

Solution is simple, is to have the same name. Either install PG with **lisk_sdk_db** or run the command to start the container with **lisk_db** instead

So, **one** of these commands:

#### On the Postgres instalation section:

```bash
docker run --name lisk_sdk_db -p 5432:5432 -e POSTGRES_USER=lisk -e POSTGRES_PASSWORD=password -e POSTGRES_DB=lisk_dev -d postgres:10
```

#### On the tutorial section:

```bash
docker exec -ti lisk_db psql -h localhost -U lisk -d postgres -c "DROP DATABASE lisk_dev"
```

Tschakki:
Closes #727

### DIFF
--- a/modules/ROOT/pages/tutorials/hello-world.adoc
+++ b/modules/ROOT/pages/tutorials/hello-world.adoc
@@ -112,8 +112,8 @@ If you have installed xref:{url_setup_postgresql}[Postgres with Docker], it is n
 
 [source,bash]
 ----
-docker exec -ti lisk_sdk_db psql -h localhost -U lisk -d postgres -c "DROP DATABASE lisk_dev"
-docker exec -ti lisk_sdk_db psql -h localhost -U lisk -d postgres -c "CREATE DATABASE lisk_dev OWNER lisk"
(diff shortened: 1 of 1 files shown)


itstheandre (approved):
### Its exactly this. I believe now it should be totally unified 👍

PR #756, opened 2020-09-28
Update protocol docs according to on chain architecture changes

janhack:
Update the  introduction, blocks, consensus and network section with respect to the following:

- On-chain architecture changes, removal of transaction types and introduction of the module concept.
- New transaction terminology.
- Minor formulation improvements.

Closes #752

PR #757, opened 2020-09-29
 Update introduction, blocks, consensus and network section #756 

### LINKED ISSUE #752: Update protocol documentation to reflect on-chain architecture changes.
The protocol documentation requires some minor changes due to the changes introduced by the [on-chain architecture](https://github.com/LiskHQ/lisk-sdk/issues/5603).

janhack:
Update the introduction, blocks, consensus and network section with respect to the following:

- On-chain architecture changes, removal of transaction types and introduction of the module concept.
- New transaction terminology.
- Minor formulation improvements.

Closes #752


modules/ROOT/pages/index.adoc
```diff
@@ -30,16 +30,15 @@ The validity of a block (and all transactions within it) is determined against t
...
-
 == Overview of the Lisk protocol
 
 The Lisk protocol is the formal set of rules that defines a blockchain created with the Lisk SDK.
-The Lisk protocol defines an account-based blockchain with a fixed set of [#index-transaction_types-1]#*default transaction types*# that are used to modify the accounts state.
-The fees for those transactions are paid with the chain native token.
+The Lisk protocol defines an account-based blockchain with a set of default transaction that are used to modify the accounts state.
+The account state and transaction logic is structured into *modules*.
+A module can define new account properties and a set of transactions acting on the whole account state.
+The Lisk SDK contains four default modules: token, sequence, keys, and dpos.
```
IkerAlus: I think it is ok to explain this point here and in transactions section. However I suggest to follow here the same format (i.e., `token`) as in the transaction section.

modules/ROOT/pages/consensus-algorithm.adoc
```diff
@@ -7,25 +7,26 @@
 :url_transactions_delegate: transactions.adoc#delegate
 :url_transactions_pom: transactions.adoc#pom
 :url_transactions_vote: transactions.adoc#vote
+:url_blocks_genesis: blocks.adoc#genesis_block
```
IkerAlus: the right link handle is `blocks.adoc#genesis-block`

IkerAlus (changes requested):
Apart from the minor format fixes, the rest LGTM.

Also, a little detail in the block.png diagram, for consistency the transaction object should be updated to the new version (`asset: Bytes`)

modules/ROOT/pages/index.adoc
janhack: I am not sure about the suitable typesetting. On the one hand "token" is the actual string name of the module in the code, on the other hand we also don't use monospace when we write "transfer transaction", even though transfer was the string name of the base asset (now the name is token transfer transaction though). If I write "... is defined in the token module", I am not sure if monospace is nice.

modules/ROOT/pages/consensus-algorithm.adoc
janhack: Good catch. I fixed it.

IkerAlus (approved):
LGTM

PR #831, opened 2020-12-08
Add strong password note to passphrase encryption

IkerAlus:
In the pages _forging.adoc_ and _cryptography.adoc_

### DIFF
--- a/modules/ROOT/pages/guides/node-management/forging.adoc
+++ b/modules/ROOT/pages/guides/node-management/forging.adoc
@@ -17,6 +17,9 @@ Mona Bärenfänger <mona@lightcurve.io>
 :url_protocol_delegate_selection: protocol:consensus-algorithm.adoc#delegate_selection
 :url_protocol_blocks_punishment: protocol:consensus-algorithm.adoc#punishment
 
+:url_wikipedia_password_strength: https://en.wikipedia.org/wiki/Password_strength#Guidelines_for_strong_passwords
+
+
 Describes how to enable forging on a node as a delegate.
 
 == Registering a delegate
@@ -63,7 +66,9 @@ Please re-enter your password: ***
 
 [NOTE]
 ====
-Be sure to use a link:https://en.wikipedia.org/wiki/Password_strength#Guidelines_for_strong_passwords[strong password]. 
+Be sure to use a strong password.
+
+See the {url_wikipedia_password_strength}[Guidelines for password strength^] at Wikipedia for reference. 
 ====
 
 * Type in the passphrase followed by the password required for encryption.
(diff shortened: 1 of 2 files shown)


Tschakki (changes requested):
Maybe it would help to describe what is considered a strong password in this context. Or Add a link to a page, which is covering the topic.

IkerAlus:
> Maybe it would help to describe what is considered a strong password in this context. Or Add a link to a page, which is covering the topic.

I was planning to do so and link to [this wiki article](https://en.wikipedia.org/wiki/Password_strength) which has a very comprehensive description of the concept. But I did not see any external link in these pages so I thought maybe we prefer not to include those. If you think it is ok to have an external here I will add the mentioned link as suggested.

Tschakki:
Hmm the article is very general. It does not give me an answer what is secure in this regard. I think we should give concrete guidelines here to avoid confusion. For example if we expect to use the 12 word mnemonic for the password as well, we should mention it directly. We can still include this external link for reference. Generally, there are no places in docs where it's forbidden to add external links :)

IkerAlus:
The point here is that this is not a basic user level application where we can set strict rules, but a tool aimed to delegates and node operators. There should be not limitations or basic guidelines. If we wanted to be precise and complete regarding the password strength concept we would need to elaborate like in the wiki article, which is not the point of these doc pages. Moreover, that article has the precise guidelines for what we understand as a _strong password_ in the section "Guidelines for strong passwords". What we can do is reference to that exact section to avoid the reader to get lost in the bigger wiki article.

modules/ROOT/pages/guides/node-management/forging.adoc
```diff
@@ -61,6 +61,11 @@ Please re-enter your password: ***
 <1> Enter the secret passphrase here that needs to be encrypted.
 <2> Enter the password here that will be required to decrypt the passphrase again.
 
+[NOTE]
+====
+Be sure to use a link:https://en.wikipedia.org/wiki/Password_strength#Guidelines_for_strong_passwords[strong password].
```
Tschakki: Please always use link attributes to insert links:
On top of the page under settings:

```
:url_wikipedia_password_strength: https://en.wikipedia.org/wiki/Password_strength#Guidelines_for_strong_passwords
```

Then use the attribute in the text.
```
Be sure to use a strong password.

See the {url_wikipedia_password_strength}[Guidelines for password strength^] at Wikipedia for reference. 
```

The caret at the end will open the link in a new tab.

Tschakki (changes requested):

modules/ROOT/pages/guides/node-management/forging.adoc
IkerAlus: added, thanks!

Tschakki (approved):
Thank you!

Issue #860, opened 2020-12-15
Change `sentVotes` to `votes` in DposObject diagram


AndreasKendziorra:
The diagram for `DposObject` in the [Account](https://github.com/LiskHQ/lisk-docs/blob/docs-sdk/modules/protocol/pages/accounts.adoc) documentation contains the property `sentVotes`. However, the name of this property should be `votes`. The name `votes` is also used in the text.

PR #878, opened 2020-12-28
Fix typo and change diagram format

ricott1:
Fix typo in Account section: https://github.com/LiskHQ/lisk-docs/issues/860

Change diagram format to svg and remove unif_diagrams nested folder.


ricott1:
The diagrams size is left to default. However, the transaction schema diagrams would look nicer if rescaled from the original size. AFAIK, there is no easy way to do this with asciidoc (scale, and scaledwidth options did not work). We could still change them all one by one

MaximeGagnebin:
For reference: in the code, the array storing the votes cast by an account is called `sentVotes` (see https://github.com/LiskHQ/lisk-sdk/blob/498013319f16a52d38fe0bfdf4ecced1a1ac62d2/framework/src/modules/dpos/schema.ts#L86). This naming is used here too for consistency.

AndreasKendziorra (approved):
The change regarding 'votes' -> 'sentVotes' looks good to me. But I did not check the diagram formats.

Tschakki:
@ricott1 please use `docs-sdk` as base branch for your PR, we moved the Lisk protocol content over there

PR #1069, opened 2021-04-29
Update lisk-genesis reference - Closes #1067

### LINKED ISSUE #1067: Update lisk-genesis reference
Elaborate more on the lisk-genesis interface.

`createGenesisBlock` is most important function.

```
// List of initial accounts in the network. Minimum required is address, but can have other properties like balance. 
readonly accounts: ReadonlyArray<Partial<Account<T>> & { address: Buffer }>;
// List fo initial delegate addresses used during the bootstrap period to forge blocks
readonly initDelegates: ReadonlyArray<Buffer>;
// Account Schema for the genesis block. Contains all account asset schemas of all modules which are part of the blockchain application.
readonly accountAssetSchemas: accountAssetSchemas;
// Number of rounds for bootstrap period, default is 3
readonly initRounds?: number;
// Optional. Height of the genesis block
readonly height?: number;
// Optional. Timestamp of the genesis block.
readonly timestamp?: number;
// Optional. previous block ID. Can be used for example in case of regenesis.
readonly previousBlockID?: Buffer;
```

See: https://github.com/LiskHQ/lisk-sdk/blob/development/elements/lisk-genesis/src/types.ts

Tschakki:
Closes #1067

### DIFF
--- a/modules/ROOT/pages/references/lisk-elements/genesis.adoc
+++ b/modules/ROOT/pages/references/lisk-elements/genesis.adoc
@@ -30,9 +30,10 @@ const { TokenModule, DPoSModule, genesis, configDevnet } = require('lisk-sdk');
 // Get account schemas of the modules which are registered with the application
 const token = new TokenModule(configDevnet.genesisConfig).accountSchema;
 const dpos = new DPoSModule(configDevnet.genesisConfig).accountSchema;
-// Add fieldNumber property, to avoid errors
+// Add fieldNumber starting from 2. Field number 1 is assigned to address of the account
 dpos.fieldNumber = 2;
-token.fieldNumber = 9;
+token.fieldNumber = 3;
+
 // Pack all account schemas in one big object
 const accountAssetSchemas = {
   token,
--- a/modules/ROOT/pages/references/lisk-elements/genesis.adoc
+++ b/modules/ROOT/pages/references/lisk-elements/genesis.adoc
@@ -31,8 +31,8 @@ const { TokenModule, DPoSModule, genesis, configDevnet } = require('lisk-sdk');
 const token = new TokenModule(configDevnet.genesisConfig).accountSchema;
 const dpos = new DPoSModule(configDevnet.genesisConfig).accountSchema;
 // Add fieldNumber starting from 2. Field number 1 is assigned to address of the account
-dpos.fieldNumber = 2;
-token.fieldNumber = 3;
+token.fieldNumber = 2;
+dpos.fieldNumber = 3;
 
 // Pack all account schemas in one big object
 const accountAssetSchemas = {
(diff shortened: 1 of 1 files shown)


modules/ROOT/pages/references/lisk-elements/genesis.adoc
```diff
@@ -25,35 +25,114 @@ npm update @liskhq/lisk-genesis
...
+			balance: {
+				fieldNumber: 1,
+				dataType: 'uint64',
+			},
+		},
+		default: {
+			balance: BigInt(0),
+		},
+	}
+	// dpos: {...}
```
nazarhussain: `DPOS` is kind of required schema, you can't skip it. Because without a initDelgates list you can't create a genesis block and for that we must need to include the `dpos` module schema.

modules/ROOT/pages/references/lisk-elements/genesis.adoc
```diff
@@ -25,35 +25,114 @@ npm update @liskhq/lisk-genesis
...
+		properties: {
+			balance: {
+				fieldNumber: 1,
+				dataType: 'uint64',
+			},
+		},
+		default: {
+			balance: BigInt(0),
+		},
+	}
```
nazarhussain: People may get confused about this hardcoded schema. We should refer it directly to the framework. 

```
import {TokenModule} from 'lisk-sdk';

const accountAssetSchemas = { token: new TokenModule().accountSchema }
```

Must test it before, there my be some parameters required to create TokenModule object.

modules/ROOT/pages/references/lisk-elements/genesis.adoc
```diff
@@ -25,35 +25,114 @@ npm update @liskhq/lisk-genesis
...
-	height: 5,
-	timestamp: 1591873718,
-	previousBlockID: Buffer.from(
-		'454690a1c37838326007519a7ce1c8a6a495df50898f1ebd69d22fbcedf9689a',
-		'hex',
-	),
-	initDelegates: validDelegateAccounts.map(a => a.address),
-	accounts: [...validAccounts, ...validDelegateAccounts],
+	initDelegates: delegates.map(a => a.address),
+	accounts,
```
nazarhussain: Accounts must contains all the account including the delegate accounts. In here you are passing only normal accounts.

nazarhussain (changes requested):

modules/ROOT/pages/references/lisk-elements/genesis.adoc
Tschakki: You seem to need the genesis config to create instances of all default modules. How to best retrieve it from the application?

Tschakki: the genesis config is needed as parameter.
I have now made changes that use the genesis config of the `configDevnet` sample as parameter to get the corresponding account schemas.

There is one issue with the account schemas, `createGenesisBlock()` does not accept them without that a new property `fieldNumber` is added to the account schemas.

I added it manually as a workaround, but it is kind of weird like this. Is that a known issue?

nazarhussain (approved):

PR #1100, opened 2021-06-01
initial lisk core snap documentation

Nazgolze:
A first pass at updating the docs to account for `snap` installs


Tschakki (approved):
Thanks!

PR #1118, opened 2021-06-16
Glossary updates in SDK docs v5 - Closes #1117

### LINKED ISSUE #1117: Glossary update in SDK docs v5
Add in a brief explanation regarding Faucets into the Glossary

chris529:
Faucet and Validator descriptions added in Glossary.

Closes #1117

### DIFF
--- a/modules/ROOT/pages/references/glossary.adoc
+++ b/modules/ROOT/pages/references/glossary.adoc
@@ -458,10 +458,6 @@ In other words it can be thought of as an indivisible chunk of a token, which is
 
 == V
 
-=== Validator
-
-A validator can be described as a user who is authorized to secure the network. In a DPoS system there are a fixed number of validators that are able to vote on who will validate transactions. The validators who receive the highest number of votes can then become delegates.
-
 === Voter
 
 An account deploying their LSK to vote for their favourite delegates.
--- a/modules/ROOT/pages/references/glossary.adoc
+++ b/modules/ROOT/pages/references/glossary.adoc
@@ -160,8 +160,7 @@ Ed25519 signatures are elliptic-curve signatures, carefully engineered at severa
 
 === Faucet
 
-Lisk has a faucet plugin for the Lisk Framework that provides a user interface to receive an certain amount of tokens on request.
-In addition, the Lisk faucet plugin has the ability to send tokens to different accounts from a faucet.
+A Faucet is generally defined in the blockchain world as a tool whereby users can receive a certain amount of tokens for free. For Lisk blockchain apps, the faucet plugin can be registered with the application to provide a basic faucet, which is useful during the development phase, and also for proof-of-concept blockchain applications.
 
 === FireHOL
 
--- a/modules/ROOT/pages/references/glossary.adoc
+++ b/modules/ROOT/pages/references/glossary.adoc
@@ -160,7 +160,7 @@ Ed25519 signatures are elliptic-curve signatures, carefully engineered at severa
 
 === Faucet
 
-A Faucet is generally defined in the blockchain world as an vehicle for distributing small amounts of tokens. Lisk has a faucet plugin for the Lisk Framework that provides a user interface to receive an certain amount of tokens on request.
+Lisk has a faucet plugin for the Lisk Framework that provides a user interface to receive an certain amount of tokens on request.
 In addition, the Lisk faucet plugin has the ability to send tokens to different accounts from a faucet.
 
 === FireHOL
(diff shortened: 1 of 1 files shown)


modules/ROOT/pages/references/glossary.adoc
```diff
@@ -158,6 +158,11 @@ Ed25519 signatures are elliptic-curve signatures, carefully engineered at severa
 
 == F
 
+=== Faucet
+
+A Faucet is generally defined in the blockchain world as an vehicle for distributing small amounts of tokens. Lisk has a faucet plugin for the Lisk Framework that provides a user interface to receive an certain amount of tokens on request.
+In addition, the Lisk faucet plugin has the ability to send tokens to different accounts from a faucet.
```
Tschakki: This sentence sounds weird. What exactly are you referring to?

modules/ROOT/pages/references/glossary.adoc
```diff
@@ -454,6 +459,10 @@ In other words it can be thought of as an indivisible chunk of a token, which is
 
 == V
 
+=== Validator
+
+The validator is essentially a validation library according to the Lisk Protocol. The @liskhq/lisk-validator package is a wrapper of validation packages, that includes custom validation utilities related to the Lisk Protocol.
```
Tschakki: I don't think this is the validator wee want to explain here. This should refer to validators in the DPoS system. Please clarify, if and how we want to use the word validator at all, because we always speak of delegates in the documentation at the moment.

Tschakki (changes requested):

modules/ROOT/pages/references/glossary.adoc
chris529: Understood, so I have removed the first sentence. Do you think the remaining description is OK now?

modules/ROOT/pages/references/glossary.adoc
chris529: Yes, well I have tried to reword it accordingly...Do you think this is now more relevant...?

modules/ROOT/pages/references/glossary.adoc
Tschakki: Hm, I was not really referring to the first sentence, as you see my comment is under the third sentence.

For me it is confusing to say "in addition", as it is not really something additionally, as this is already implied by the second sentence.

modules/ROOT/pages/references/glossary.adoc
```diff
@@ -160,7 +160,7 @@ Ed25519 signatures are elliptic-curve signatures, carefully engineered at severa
 
 === Faucet
 
-A Faucet is generally defined in the blockchain world as an vehicle for distributing small amounts of tokens. Lisk has a faucet plugin for the Lisk Framework that provides a user interface to receive an certain amount of tokens on request.
+Lisk has a faucet plugin for the Lisk Framework that provides a user interface to receive an certain amount of tokens on request.
```
Tschakki: This I would keep or update to something like this

```suggestion
A Faucet is generally defined in the blockchain world as tool where users can receive a certain amount of tokens for free. For Lisk blockchain apps, the faucet plugin can be registered with the application to provide a basic faucet, which is useful during the development and for proof-of-concept blockchain applications.
```

thats my suggestion :)

modules/ROOT/pages/references/glossary.adoc
```diff
@@ -461,7 +461,7 @@ In other words it can be thought of as an indivisible chunk of a token, which is
 
 === Validator
 
-The validator is essentially a validation library according to the Lisk Protocol. The @liskhq/lisk-validator package is a wrapper of validation packages, that includes custom validation utilities related to the Lisk Protocol.
+A validator can be described as a user who is authorized to secure the network. In a DPoS system there are a fixed number of validators that are able to vote on who will validate transactions. The validators who receive the highest number of votes can then become delegates.
```
Tschakki: We should clarify if and how we want to use the word validator in the docs.
As mentioned, we are now only referring to them as delegates, so I wonder why add this entry to the glossary.

Tschakki (changes requested):

modules/ROOT/pages/references/glossary.adoc
chris529: Hey Mona, It looks very good, although I just noticed a couple of minor errors here so I have slightly adapted it and cut and pasted the slightly modified version following into my WS branch: 
"A Faucet is generally defined in the blockchain world as a tool whereby users can receive a certain amount of tokens for free. For Lisk blockchain apps, the faucet plugin can be registered with the application to provide a basic faucet, which is useful during the development phase, and also for proof-of-concept blockchain applications".

modules/ROOT/pages/references/glossary.adoc
chris529: Agreed..After our discussion, I did some research with a view to possibly adding in the term'Validators' under the ''Delegate'' description. However, I am now of the opinion that actually we don't need to mention this term yet (at least not at his point anyhow). so if you are in agreement, I have just deleted the Validator description.

Tschakki (approved):
Thanks!

PR #1131, opened 2021-06-27
Update plugin.adoc

newway-anshul:
Instead of having "The command generate: asset expects 1 argument:" we should have "The command generate: plugin expects 1 argument:".


Tschakki (approved):
Thanks!

PR #1143, opened 2021-07-08
Change 'API (Betanet)' to 'API (Testnet)'

Nazgolze:


Tschakki (changes requested):
Good change, but the URL on line 5 needs update as well, should be version 5.1.1

Tschakki (approved):
Thank you!

PR #1162, opened 2021-07-31
Update keys-module.adoc - wrong module name

krzyhub:
Wrong module name used in the description. "token module" used, but should be "keys module".


Tschakki (approved):
Thank you!

PR #1168, opened 2021-08-12
More info about EXCHANGERATESAPI_IO_API_KEY

adamant-al:
Hi,
I am confused about `EXCHANGERATESAPI_IO_API_KEY` option.
First, `/market/prices` works even with no `EXCHANGERATESAPI_IO_API_KEY` option set. Where this data comes from then?
Second, what benefits makes setting `EXCHANGERATESAPI_IO_API_KEY` option?
Third, what pricing plan Lisk Service expects from https://exchangeratesapi.io/pricing/?

### DIFF
--- a/modules/ROOT/pages/references/configuration.adoc
+++ b/modules/ROOT/pages/references/configuration.adoc
@@ -197,7 +197,7 @@ SERVICE_MARKET_TARGET_PAIRS=LSK_BTC,LSK_EUR,BTC_CHF
 
 # Optional API key for https://exchangeratesapi.io/
 # Used to calculate Fiat exchange rates
-EXCHANGERATESAPI_IO_API_KEY=EXCHANGERATESAPI_IO_API_KEY
+`EXCHANGERATESAPI_IO_API_KEY=EXCHANGERATESAPI_IO_API_KEY`
 
 # Local Redis cache for the Market microservice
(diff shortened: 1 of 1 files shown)


sameersubudhi:
@adamant-al `/market/prices` endpoint uses multiple sources to fetch the raw data of which `https://exchangeratesapi.io` is one that is primarily responsible for the fiat exchange rates. With a valid `EXCHANGERATESAPI_IO_API_KEY`, the `/market/prices` endpoint would then respond with additional data, specifically the exchange rates for various cryptocurrencies in other fiat currencies. The `free plan` would suffice for the purpose of Lisk Service.

adamant-al:
> endpoint uses multiple sources to fetch the raw data

I better want to know where my server make requests to. And have a possibility to disable them, if I don't need market rates data.

> of which `https://exchangeratesapi.io` is one that is primarily responsible for the fiat exchange rates. With a valid `EXCHANGERATESAPI_IO_API_KEY`, the `/market/prices` endpoint would then respond with additional data, specifically the exchange rates for various cryptocurrencies in other fiat currencies. The `free plan` would suffice for the purpose of Lisk Service.

Thank you. That should be mentioned in the docs. With specific info, what is `exchange rates for various cryptocurrencies in other fiat currencies`, as I see, I can get fiat rates without this key.

PR #1320, opened 2021-12-16
Improve schema explanations - Closes #1236

### LINKED ISSUE #1236: Improve schema explanations
### What
https://lisk.com/documentation/lisk-sdk/advanced-explanations/schemas.html

### How
On the schemas page, 
- provide a deeper explanation of the JSON schema being used, and how and why it differs from the official JSON schema.
- provide examples for different schema errors and how to solve them
- mention to not use `default` property for asset schemas.
- what data types are used for which data, especially adresses:
They should be stored in bytes in the DB to minimize the size: size is different between Buffer vs hex string. hex is double size if you want to represent a binary data
- Explain how to decode a transaction - show examples for all different options (api client & lisk-codec)

On asset intro & guide,
- mention that the schema must have at least one property required

### Additional information
The $id property is straightly inherited from json-schema . So what ever guidelines specified in json-schema also applies here. You can read more on https://json-schema.org/understanding-json-schema/structuring.html#the-id-property

In general 
- Use unique ids across the system 
- Use path like format (for easy readability not a requirement)
- Use one fixed identifier for your app in each id to not mix any schema other registered schema.

Add note about assets to intro and/or guides:
"To ensure that this does not create future issues, we suggest to specify every property of custom assets as required. In Lisk JSON schemas, this is done by including the name of the property in the array value of the required keyword of the object. This guarantees that unset fields are detected on the validation layer and that invalid binaries are not generated."
 
Schemas we have are inspired from JSON-Schema from where we inherited this "required" attribute. If you are using schema for validation purpose feel free to use "required" to put any property as required or not. But if you are using that schema for serialization (which is our case for account and transaction) then I recommend put all properties as required to guarantee the uniqueness of encoding as referred in the link I shared above.

---
From script, u can try creating api client and 
```
client.transaction.decode('08021000181120c096b1022a209b11208598fe779e6ae49d5c7ee8cc5d334f52e3727822f38bd74e6ce57af470321b08afa312146cc9bedb09b6bb6a73d9203501207f9546a5c06e1a003a40f0db16ec48d4d84799bd90e4646f73b2db74dcdeb0f50173d307dd38c0e621ce9b728e7bc8ff8f8c1f8268cd301060c4a35e87d16f4ecc411bf10c3585d4490e')
```
something like this should work
```
codec.decode(transactionSchema, Buffer.from('08021000181120c096b1022a209b11208598fe779e6ae49d5c7ee8cc5d334f52e3727822f38bd74e6ce57af470321b08afa312146cc9bedb09b6bb6a73d9203501207f9546a5c06e1a003a40f0db16ec48d4d84799bd90e4646f73b2db74dcdeb0f50173d307dd38c0e621ce9b728e7bc8ff8f8c1f8268cd301060c4a35e87d16f4ecc411bf10c3585d4490e', 'hex'));
and the transaction schema is also in https://lisk.com/documentation/lisk-sdk/guides/node-management/signing-transactions-offline.html
```

Tschakki:
Closes #1236

### DIFF
--- a/modules/ROOT/pages/advanced-explanations/schemas.adoc
+++ b/modules/ROOT/pages/advanced-explanations/schemas.adoc
@@ -8,6 +8,7 @@ Mona Bärenfänger <mona@lightcurve.io>
 // Project URLs
 :url_introduction_modules_accountschema: introduction/modules.adoc#account-schema
 :url_introduction_modules_assetschema: introduction/modules.adoc#transaction-asset-schema
+:url_guides_decode: guides/app-development/decoding-transactions-blocks.adoc
 :url_lisk_sdk: glossary.adoc#lisk-sdk
 
 Schemas are used in various places in the Lisk SDK to encode and decode the data that is retrieved or pushed to the database.
@@ -137,7 +138,19 @@ Data that is provided by *actions* and *events* is always returned in JSON forma
 
 |===
 
+== What data types are used for which data?
+
+Addresses, blocks and transactions::
+Should be stored in bytes in the DB to minimize the size.
+A hex string is about double size the size of the corresponding Buffer / binary data.
+
 == Decoding and encoding data
+
+=== With the API client
+
+Please refer to the dedicated guide xref:{url_guides_decode}[].
+
+=== With lisk-codec
 
 To conveniently decode and encode the data structures stored on the blockchain, such as blocks, transactions, account state, and chain state, use the `codec` library.
 The library can be imported from the following NPM packages:
--- a/modules/ROOT/pages/advanced-explanations/schemas.adoc
+++ b/modules/ROOT/pages/advanced-explanations/schemas.adoc
@@ -10,8 +10,7 @@ Mona Bärenfänger <mona@lightcurve.io>
 :url_introduction_modules_assetschema: introduction/modules.adoc#transaction-asset-schema
 :url_lisk_sdk: glossary.adoc#lisk-sdk
 
-
-Schemas are used in various places in the xref:{url_lisk_sdk}[Lisk SDK] to encode and decode the data that is retrieved or pushed to the database.
+Schemas are used in various places in the Lisk SDK to encode and decode the data that is retrieved or pushed to the database.
 
 Schemas are specifically used in the following:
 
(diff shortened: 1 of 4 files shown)


modules/ROOT/pages/advanced-explanations/schemas.adoc
```diff
@@ -27,45 +28,67 @@ NOTE: It is required to use camelCase for the key naming.
...
-The `$id` doesn't need to follow a particular format, except that it needs to be unique throughout the application.
-<2> Root type must be type `object`.
-<3> Required properties.
-<4> Properties of the schema.
-<5> Default property values.
+=== $id
+
+Unique identifier of the schema throughout the system.
+
+The $id property is straightly inherited from the JSON-schema.
```
chris529:
```suggestion
The $id property is directly inherited from the JSON-schema.
```

modules/ROOT/pages/advanced-explanations/schemas.adoc
```diff
@@ -27,45 +28,67 @@ NOTE: It is required to use camelCase for the key naming.
...
+You can read more about the `id` property on {url_json_schema_id}[^].
+
+In general:
```
chris529:
```suggestion
In general, adhere to the following criteria:
```

modules/ROOT/pages/advanced-explanations/schemas.adoc
```diff
@@ -27,45 +28,67 @@ NOTE: It is required to use camelCase for the key naming.
...
+
+* Use unique ids across the system
```
chris529:
```suggestion
* Use unique ids across the system.
```

modules/ROOT/pages/advanced-explanations/schemas.adoc
```diff
@@ -27,45 +28,67 @@ NOTE: It is required to use camelCase for the key naming.
...
+* Use path like format (for easy readability not a requirement)
```
chris529:
```suggestion
* Use path like format for easy readability, (it is not an actual requirement).
```

modules/ROOT/pages/advanced-explanations/schemas.adoc
```diff
@@ -27,45 +28,67 @@ NOTE: It is required to use camelCase for the key naming.
...
+* Use one fixed identifier for your app in each id to not mix any schema other registered schema.
```
chris529:
```suggestion
* in order to ensure no mixing of any schema with other registered schemas occurs, use one fixed identifier for your app in each id.
```

modules/ROOT/pages/advanced-explanations/schemas.adoc
```diff
@@ -27,45 +28,67 @@ NOTE: It is required to use camelCase for the key naming.
...
+
+=== properties
+
+Properties of the schema.
+
+=== default
+
+Default property values.
+
+IMPORTANT: Don't define any `default` properties for asset schemas.
```
chris529:
```suggestion
IMPORTANT: Do not define any `default` properties for asset schemas.
```

modules/ROOT/pages/advanced-explanations/schemas.adoc
```diff
@@ -173,9 +202,12 @@ await stateStore.chain.set(
 
 == Converting between formats
 
-Some data is stored as `Buffer` in the database, but in other parts of the application, the data is expected as a hexadecimal string.
+Account addresses are stored as binary data in the database.
+But in other parts of the application, the data is expected as a hexadecimal string, which is more human-readable.
```
chris529:
```suggestion
However, in other parts of the application, the data is expected as a hexadecimal string, which is more human-readable.
```

modules/ROOT/pages/advanced-explanations/schemas.adoc
```diff
@@ -173,9 +202,12 @@ await stateStore.chain.set(
...
+
+On the other hand, the node API returns the data like blocks and transactions always in hex string, but to decode transactions or blocks, they need to be in `Buffer` format.
```
chris529:
```suggestion
On the other hand, the node API always returns the data such as blocks and transactions in hex string. However, please be aware that to decode transactions or blocks, they need to be in `Buffer` format.
```

modules/ROOT/pages/introduction/modules.adoc
```diff
@@ -333,7 +333,7 @@ image::intro/asset.png[]
 
 The asset schema defines the *custom data structure of the transaction*.
 
-It defines which properties can be included, if they are optional or required, and also which data types are to expect.
+It defines which properties are required in the transaction asset, and also which data types are to expect.
```
chris529:
```suggestion
It defines which properties are required in the transaction asset, and also which data types are to be expected.
```

chris529 (commented):
Looks good, just a few minor points 👍

modules/ROOT/pages/advanced-explanations/schemas.adoc
Tschakki:
```suggestion
* Use unique IDs across the system.
```

modules/ROOT/pages/advanced-explanations/schemas.adoc
Tschakki:
```suggestion
* in order to ensure no mixing of any schema with other registered schemas occurs, use one fixed identifier for your app in each ID.
* ```

mehmetegemen (approved):

PR #1322, opened 2021-12-28
Add RPC example outputs - Closes #870

### LINKED ISSUE #870: Add missing example outputs to the RPC endpoints page
The RPC endpoints page has some missing example outputs for events and actions, that need to be added.

### Affected versions
v5

Tschakki:
Closes #870

### DIFF
--- a/modules/ROOT/pages/advanced-explanations/rpc-endpoints.adoc
+++ b/modules/ROOT/pages/advanced-explanations/rpc-endpoints.adoc
@@ -73,14 +73,20 @@ Fired when the network has at least one outbound connection.
 
 === `app:network:event`
 Fired when the application receives P2P event from the network.
-////
+
 ==== Returns
 .Example output
 [source,json]
 ----
-
+{
+  "event":"postBlock",
+  "data":
+    {
+      "block":"0ad301080210fc88ad8e06189e192220454717ffe85ae4877656962acb065eb316dcd9118a024f9f70a5c335368f03f52a20e3b0c44298fc1c149afbf4c8996fb92427ae41e4649b934ca495991b7852b8553220fd0ec19ef02054efd85dd2693ced1e28654f6e29638d5bb5de75f1dd75ab179a3880cab5ee01421808a51810bf181a10c43fe38f9ecdcf87a52615111054a1ff4a408e1897e9535be3be53f68ed7604af51852530eee6acdb7ac0ab647a2f14a23ad326253a2f7b44041efee7dbf776c1687c945115aaf83ee1a03167707c9a9840f"
+    }
+}
 ----
-////
+
 
 === `app:transaction:new`
 Fired when the node receives a new transaction.
@@ -409,14 +415,27 @@ Returns all connected peers.
 ==== Input
 none
 
-////
 ==== Returns
 .Example output
 [source,json]
 ----
-[]
+[
+  {
+    "networkIdentifier": "97fe70c3289c3c4cd310891952ab457dd86e345d37a5144aa96d2ed8892d7c82",
+    "nonce": "09e2fc804dc18386",
+    "networkVersion": "1.0",
+    "options": {
+      "height": 3240,
+      "maxHeightPrevoted": 3143,
+      "blockVersion": 2,
+      "lastBlockID": "fc108ac8e2a39eb00ba0dc9bb2158f3e1b77b0c2dfd6f014ebcf4a5d6092f6de"
+    },
+    "ipAddress": "127.0.0.1",
+    "port": 5005,
+    "peerId": "127.0.0.1:5005"
+  }
+]
 ----
-////
 
 === `app:getDisconnectedPeers`
 Returns all disconnected peers
@@ -424,14 +443,24 @@ Returns all disconnected peers
 ==== Input
 none
 
-////
 ==== Returns
 .Example output
 [source,json]
 ----
-[]
+[
+  {
+    "networkIdentifier": "97fe70c3289c3c4cd310891952ab457dd86e345d37a5144aa96d2ed8892d7c82",
+    "nonce": "09e2fc804dc18386",
+    "networkVersion": "1.0",
+    "options": {},
+    "ipAddress": "127.0.0.1",
+    "port": 5005,
+    "peerId": "127.0.0.1:5005"
+  }
+]
+
+
 ----
-////
 
 === `app:getForgingStatus`
 Gets information about the forging status of the node.
--- a/modules/ROOT/pages/advanced-explanations/rpc-endpoints.adoc
+++ b/modules/ROOT/pages/advanced-explanations/rpc-endpoints.adoc
@@ -43,29 +43,22 @@ To see a full list of all available actions and events of your blockchain applic
 === `app:ready`
 Fired when the application starts.
 
-////
 ==== Returns
 .Example output
 [source,json]
 ----
-{
-   "module":"app",
-   "name":"ready",
-   "data":{}
-}
+{}
 ----
-////
 
 === `app:shutdown`
 Fired when the application stops.
-////
+
 ==== Returns
 .Example output
 [source,json]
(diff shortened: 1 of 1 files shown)


modules/ROOT/pages/advanced-explanations/rpc-endpoints.adoc
```diff
@@ -450,29 +415,52 @@ Returns all connected peers.
...
+      "height": 3240,
+      "maxHeightPrevoted": 3143,
+      "blockVersion": 2,
+      "lastBlockID": "fc108ac8e2a39eb00ba0dc9bb2158f3e1b77b0c2dfd6f014ebcf4a5d6092f6de"
+    },
+    "ipAddress": "127.0.0.1",
+    "port": 5005,
+    "peerId": "127.0.0.1:5005"
+  }
+]
```
Incede, line 437: Question: Would you prefer the output format like for `app:chain:validators:change` or as it is now ?

something like 
```
{ peers: [ { }, { } .... { } ] }
```

And maybe add an example output with multiple peers.

modules/ROOT/pages/advanced-explanations/rpc-endpoints.adoc
```diff
@@ -450,29 +415,52 @@ Returns all connected peers.
...
 ==== Input
 none
 
-////
 ==== Returns
 .Example output
 [source,json]
 ----
-[]
+[
```
Incede, line 450: same, multiple peer output example if possible

modules/ROOT/pages/advanced-explanations/rpc-endpoints.adoc
```diff
@@ -521,19 +509,35 @@ Height of the previously last forged block.
...
 ==== Returns
 
-.Example output
+.Example output after disabling forging
 [source,json]
 ----
-
+{
+  "address": "d6a79b4f59f7da027405277b55783ba7dd255aa9"
+}
```
Incede, line 520: What is this address in the example conveying here ?

modules/ROOT/pages/advanced-explanations/rpc-endpoints.adoc
```diff
@@ -521,19 +509,35 @@ Height of the previously last forged block.
...
 ----
-////
+
 
 === `app:getAccount`
 Gets information about an account based on its address.
 
+==== Example
+
+[source,js]
```
Incede, line 529: Are you adding source for only selected examples ?

modules/ROOT/pages/advanced-explanations/rpc-endpoints.adoc
```diff
@@ -567,6 +595,7 @@ Gets information about multiple accounts based on their address.
 [source,json]
 ----
 [
+  "0a14d6a79b4f59f7da027405277b55783ba7dd255aa912070880d6d2c881031a020800220208002a190a170a0c64656c65676174655f313031180120ac1628003000c23e0e0a0c48656c6c6f20576f726c6421",
```
Incede, line 598: It does not say here if it Is input or output

Incede (changes requested):

modules/ROOT/pages/advanced-explanations/rpc-endpoints.adoc
Tschakki: Can you explain more? I am not sure I understand what you mean

modules/ROOT/pages/advanced-explanations/rpc-endpoints.adoc
Tschakki: I don't have a strong opinion here, either is fine for me I guess.

About multiple outputs: It is a little cumbersome to generate these, and I don't think it is adding new information, so I would prefer to keep it as is.

modules/ROOT/pages/advanced-explanations/rpc-endpoints.adoc
Tschakki: See my reply above

modules/ROOT/pages/advanced-explanations/rpc-endpoints.adoc
Tschakki: This is the data that is returned when disabling forging via RPC action (see label of the snippet)

modules/ROOT/pages/advanced-explanations/rpc-endpoints.adoc
Tschakki: Yes it does, in the label above the snippet.

Incede: ah yes! It was truncated in Github in default view

modules/ROOT/pages/advanced-explanations/rpc-endpoints.adoc
Incede: hmm yea ik but I am trying to understand what does it mean... How is it related to `updateForgingStatus` ?

modules/ROOT/pages/advanced-explanations/rpc-endpoints.adoc
Incede: I saw in some places the source code for some examples was removed (L169, L235) or not present in others.

modules/ROOT/pages/advanced-explanations/rpc-endpoints.adoc
Tschakki: it is an example of what it returns

modules/ROOT/pages/advanced-explanations/rpc-endpoints.adoc
Tschakki: I wanted to include one example of decoding an account, which I think is missing in the docs atm.
Probably there should be a guide covering this, but for now I added a simple example here.

The other two examples were kind of old and covered elsewhere in the docs, so I removed the examples for them.

Incede: I'd suggest using a consistent approach for providing source code with examples (either do for all or none) but in the end it's up to you.

modules/ROOT/pages/advanced-explanations/rpc-endpoints.adoc
Incede: hmm idk if it's the right behavior but returning only an address as the output of `app:updateForgingStatus` doesn't make sense for me.

Tschakki: Maybe you can verify it yourself, for me this was exactly what was returned when disabling forging.
If you don't like what it returns, I suggest to create a new issue for it on the lisk-sdk repo.
I agree personally, more info, and maybe a general more standardized format for RPC replies would be great.
Also for `app:ready` and `app:shutdown` it only returned `{}` / undefined when the event was fired.

modules/ROOT/pages/advanced-explanations/rpc-endpoints.adoc
Tschakki: For now it is OK like this, but if I create the guide, I will consider remove it here for consistency 👍

modules/ROOT/pages/advanced-explanations/rpc-endpoints.adoc
Incede: Thanks for the feedback. Created an [issue](https://github.com/LiskHQ/lisk-sdk/issues/6967). Would be nice if you could also specify details what you expect in terms of "more information and standardized format" in the issue itself.

Incede (approved):

modules/ROOT/pages/advanced-explanations/rpc-endpoints.adoc
Tschakki: Thanks!

PR #1376, opened 2022-01-27
Update top nav - Closes #1351

### LINKED ISSUE #1351: Update the top navigation
## Which content is outdated?
Top navigation of the docs UI

image

Epic: #1294

Tschakki:
Closes #1351


chris529 (approved):
LGTM! 👍

shuse2 (changes requested):
Few comments for design
* Header size should be 107px
* Padding on left/right is 40px
* Top padding is 30px and bottom is 25px
* hovar style on the right link seems to be wrong
* Logo and `Docs` should be closer

Additionally, 
* Search bar style is added later?
* Dark mode we can add it later. let's remove it for now
* Product link is missing but it will be added later?

Tschakki:
@shuse2 thanks for your comments! I tried to address them as follows:

- The header height is slightely lower, but imho it is high enough, I wouldn't increase the height further unless the space would be really required for the top nav.
- I increased the padding to 32px now. As padding CSS values are defined relative (em & rem) it is a little difficult to stick to exact pixels, but I think it looks good like this. Tell me if you would like the padding to be bigger, then I will update it again.
- top/bottom padding are now 32px as well, imho it looks good symmetrically, let me know if I should change it.
- What exactly looks wrong for you in the hover style? Color, or something else?
- Removed white space between Lisk and Docs logos.

---

- Added now round border for the search bar, let me know if there is anything else to adjust.
- I would like release the new UI together with the dark mode, to have a bigger wow effect :) So for now we can keep it like this, it will be updated with the dark mode as a last step of the design before releasing it.
- As you can already see on gh-pages, where I already include the PR for updating toolbar, I moved the product and version selection to the toolbar. Imho, this is a much better solution, I would really like to keep it this way. That's why I removed the product links from the top nav, felt too redundant for me, and leaves more space for other links in the top nav.

shuse2:
* For the design, please discuss with @ann-vashchenko , and adjust the design first since figma is source of truth.
* For hover specifically, there is docs design library, so please refer to that style.
* If the search bar is included in the PR, then the style should follow the design. ie) icon, font, placeholders, size etc
* Product link was on the top right because it's important and we wanted to put it on the visible place. I think it was good on design as well

Let's try to keep the design as it is designed, and if we want to change, let's discuss and apply on the design first

shuse2 (approved):

PR #1384, opened 2022-02-08
Update next & previous link styles - Closes #1357

### LINKED ISSUE #1357: Update page prv/nxt navigation
## Which content is outdated?
Update the next and previous page links.

image

Epic: #1294

Tschakki:
Closes #1357


ui/src/css/extensions/common-custom.css
```diff
@@ -406,6 +406,28 @@ pre {
   align-self: center;
 }
 
+nav.pagination a {
+  font-weight: 400;
```
shuse2, line 410: font-size should be 16px

ui/src/css/extensions/common-custom.css
```diff
@@ -406,6 +406,28 @@ pre {
...
+}
+
+nav.pagination .next a::after {
+  content: "";
```
shuse2, line 415: this is not taking effect?

shuse2 (changes requested):
I think there should be 30px between boarder and the next/prev

ui/src/css/extensions/common-custom.css
Tschakki: for me it does 🤔

shuse2: Screenshot 2022-02-09 at 09 42 44

This one?

Tschakki: I discussed with Anna to keep the default design of the links with some small adjustments, the `content: ""` removes the old chevron icon because it is not needed anymore, it's replaced with ours. It is not affecting the texts `Next` or `Prev`.

shuse2 (approved):

PR #1388, opened 2022-02-09
Update table style - Closes #1363

### LINKED ISSUE #1363: Update style of tables
## Which content is outdated?
Update the style of tables in the documentation

image

Epic: #1294

Tschakki:
Closes #1363

### DIFF
--- a/ui/src/css/extensions/common-custom.css
+++ b/ui/src/css/extensions/common-custom.css
@@ -632,3 +632,35 @@ footer.footer {
   padding-top: 0.5rem;
   list-style: disc;
 }
+
+/* .doc table.tableblock, .doc table.tableblock > * > tr > * {
+  border-radius: 20px;
+}
+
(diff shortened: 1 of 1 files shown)


shuse2 (changes requested):
I think the borders on the image should be hidden / thinner
Screenshot 2022-02-09 at 18 21 54

Tschakki:
Decreased border width of the head bottom. Other borders are already 1px, which is smallest size. The borders should be visible in the table in my opinion.

shuse2:
@Tschakki i think the border between header and content is still thicker than it should.

for the vertical border, i think it's better to ask @ann-vashchenko ? IMO, it usually obvious with space, so not really necessary

Tschakki:
nvm, I did the requested changes 👍

shuse2 (approved):

PR #1421, opened 2022-03-07
Guides improvements - Closes #1409

### LINKED ISSUE #1409: SDK docs improvements
## Description

Check & fix the following issues in the SDK docs:

- the Hello World tutorial: asset is named createHello (with the file create_hello_asset.ts), but in the module code there is require(‘./assets/hello_asset’). Is that an inconsistency? I think, the application should not compile with create_hello_asset.ts.
Lisk-sdk tutorial, creating a module asset, the asset schema. The numbering in the code snippet doesn’t correspond to the numbering after the code.
- module registration: in modules.ts, the function registerModules has a parameter named _app, but a module attempts to be registered by app.registerModule. It doesn’t run out of the box.
- afterTransactionApply: variable _input is not defined (I think, _input = {transaction, stateStore, reducerHandler}). It should be removed.
- Creating a plugin, the end of section 3. It ends with: “For example,“, the example is missing.
- Configuring a blockchain app: the code snippet with an example of custom genesisConfig. There are 31+5 delegates, but bftThreshold is 78, does it make sense?
- Setting a custom configuration: “bin/run start --config my_custom_config.json --override-config”, no such flag —override-config, it’s —overwrite-config.
- latest_hello_plugin (the code on github), the variable _hello that stores the last message is never defined, so should throw an error?

Tschakki:
Closes #1409


sergeyshemyakov (approved):
Looks good from my side!

Incede (approved):

PR #1434, opened 2022-03-21
Fix unit test - Closes #1431

### LINKED ISSUE #1431: Test does not pass in test guide
### Description
The test for the action `amountOfHellos` in the Hello module is failing.

```
Expected: 13
Received: 0
```

Reason: the action/endpoint is not using state store. the test data needs to be inserted in to the data access, or use jest.spyOn to return the mocked data.

## Solution

Update the code snippet with the fixed example from lisk-sdk-examples repo: https://github.com/LiskHQ/lisk-sdk-examples/blob/development/guides/04-plugin/hello_app/test/unit/modules/hello/hello_module.spec.ts#L114-L118

Tschakki:
Closes #1431


Incede (approved):
I think the new test setup doesn't add much utility. Approving in case you still think it's necessary.

chris529 (approved):
LGTM! 👍

PR #1436, opened 2022-03-24
Update images for dark theme in Lisk docs - Partially closes  #1408

### LINKED ISSUE #1408: Update images for dark theme
## Which content is outdated?
Most images in the documentation look only good for the light theme.
Some even are not entirely readable in the dark theme.

The images should be updated to be readable and fitting for both, dark & light theme.

- Update images with transparent background
- Make sure the test is readable on dark and light background
- Review and adjust image sizes if necessary

### Tasks

- [x] General images
- [x] Docs images
- [x] SDK images
- [x] Core images
- [x] Service images

chris529:
Partially closes #1408


shuse2 (commented):
Some images are transparent with background, and some are white background.
Wouldn't it be more consistent to have all white background?

chris529:
@shuse2 Hi Shu, Mona and I discussed this, and decided that there were a couple of rather complex, intricate diagrams whereby it was much clearer and easier to be viewed with a white background. However, the majority should now be with a transparent background. Which particular diagrams are you referring to..? For example the P2P architecture diagram we decided to leave it with a white background.

shuse2:
I think having white background make sense and more readable, but I thought it just inconsistent to have transparent background for some image.

Tschakki (changes requested):
Overall looking really great, just a few minor comments

image Maybe for the title, it would look better without white background, same for the Module and asset diagrams. Or remove the title from the diagram, as a title is already given in the documentation, so what is to see on the diagram should be clear anyway, what are your thoughts about this?

image
Also here and the following diagrams, I think it might look better without white background behind the green text, what do you think?

image
Please remove the white border at the left of the diagram.

chris529:
Hey Mona, Agreed...You are right. This is great feedback, as it looks much better with the titles of the Plugin and Module and Assets diagrams without the white background. I quite like the titles here, but also it seemed to me maybe they were a bit too large, so I have reduced the size of them.
I will also remove the white background on the NW topology diagram and in the Devnet and Testnet diagrams. Finally, I will check the other diagrams as well, and where there is any text on a white background I will remove it. 👍

Tschakki (changes requested):
I can still see the white left border in node-network.png.

Also, I think sometimes it might be nice to reduce the image size so that the picture fits better into the content, for example here: 
image

I think it is better to display it smaller, so it is merging better with the other content.

chris529:
Hey Mona, OK thanks. I will recheck this white border issue. Yes, good point regarding the Plugin anatomy diagram...I agree it would look better here if it were reduced in size, as this is often the case with simple block-type diagrams which do not contain so much detail, as they tend to appear somewhat overpowering. No worries, I will take care of it. 👍

Tschakki (approved):
Thanks, the images are looking really great now! 👍

shuse2 (approved):

PR #1440, opened 2022-04-01
Improve navigation - Closes #1437

### LINKED ISSUE #1437: UI nav improvements
### Description

- Add border to topnav dropdown
image

- Improve styling of current page in nav
image

Tschakki:
Closes #1437

### DIFF
--- a/README.adoc
+++ b/README.adoc
@@ -1,34 +1,140 @@
-= Lisk Docs
-:imagesdir: assets
+= Repository structure and workflow
+:hide-uri-scheme:
+:idprefix:
+:toc: preamble
+// External URLs:
+:url_antora_ui: https://gitlab.com/antora/antora-ui-default
+:url_antora_ui_docs: https://docs.antora.org/antora-ui-default/
+:url_github_docs: git@github.com:LiskHQ/lisk-docs.git
+:url_antora: https://antora.org/
+:url_antora_directories: https://docs.antora.org/antora/latest/standard-directories/
+:url_docs: https://lisk.com/documentation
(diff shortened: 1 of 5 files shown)


chris529 (approved):
LGTM! 👍

shuse2:
I think font and alignment on the selected sidebar is different from the design.
  * the font size and weight should be the same
  * alignment of the letter should not change

the border style / top bar style is different from the design, but is this aligned with @ann-vashchenko ?

Tschakki:
Yes it was one of the requests for this PR to update the borders accordingly, they should be good now.

I applied the other feedback, please check again @shuse2

shuse2:
I think as per design, selected link shouldn't have blue background?

Screenshot 2022-04-11 at 12 09 30

Tschakki:
Yes I think it should, the desired design is also displayed in the issue description #1437

shuse2:
Ah sorry, it is correct it should have background blue.
The  design wise, it should be the whole raw (but i think it would be difficult?). If so, I think it should have some padding on left so the letter is not sticked to the left boarder (maybe discuss with @ann-vashchenko

Tschakki:
Yes, I think it would make it more difficult to have the background color the whole row. Some small padding would be the easier solution. Maybe we can go for the easy solution now, and then request Annas review? If she has any objections, she can include it there.

shuse2 (approved):

ann-vashchenko:
Hi @Tschakki @shuse2 What do you think if the active item will look like this?
Screenshot 2022-04-11 at 14 47 05

PR #1482, opened 2022-05-30
Fix issues in NFT tutorial - Closes #1480

### LINKED ISSUE #1480: Fix issues in NFT tutorial
### Description
1.1 Project setup
It appears that `—yes` option is removed. I can’t find it in the docs when I type `lisk init —help`, and when I run `lisk init —yes` it doesn’t act as expected.
You don’t need to install `lisk-sdk` manually, it comes with the `init` command.
You don’t need to create an `index.ts` file to bootstrap the app, it’s generated under `src/app/app.ts` automatically, and the syntax is now different.
Also, you don’t need to create a folder called `nft_module`, you just need to run
`lisk generate:module nft 1000`
and it generates the scaffold. Also consider changing from JS to TS, since mistakes are easier discovered.
In general, the tutorial creates some directories and files manually, which are in fact created automatically by the `init` command.
Nowhere in the tutorial explains what is `minPurchaseMargin` and why it should be between 1-100.

3.2 PurchaseNFT asset
The validation is done within the `apply` method instead of the `validate` method. I understand that the validation in this section are all async, but would be cool to have a line or two to explain if there's such a limitation, otherwise we should handle all the validation in the `validate` method.
Some syntaxes require slight changes to pass typescript criteria if we want to change the tutorial to typescript. This is true to all the definitions in the tutorial.

3.3 TransferNFT asset
`ReducerHandler` in the apply method is defined but never used.

4.2 The account schema
Members of the `NFTModule` class must be public . This is for the sake of consistency with the other tutorial.

5.1 The NFT plugin
The scaffold suggested for the plugin is outdated. Now the plugins reside under `src/app/plugins` . Also, you can create the base code using `lisk generate:plugin`  command automatically.

Tschakki:
Closes #1480

### DIFF
--- a/docs/modules/ROOT/pages/tutorial/nft.adoc
+++ b/docs/modules/ROOT/pages/tutorial/nft.adoc
@@ -10,6 +10,7 @@ Mona Bärenfänger <mona@lightcurve.io>
 :url_github_nft: https://github.com/LiskHQ/lisk-sdk-examples/tree/development/tutorials/nft
 :url_react_docs: https://reactjs.org/docs/getting-started.html
 // Project URLs
+:url_build_bapp: build-blockchain/create-blockchain-app.adoc#bootstrapping-a-new-blockchain-application-with-lisk-commander
 :url_references_schemas: understand-blockchain/codec-schema.adoc
 :url_statestore: lisk-sdk::references/lisk-elements/chain.adoc#state-store
 :url_references_token_module: lisk-sdk::modules/token-module.adoc
@@ -17,6 +18,9 @@ Mona Bärenfänger <mona@lightcurve.io>
 
 How to create a blockchain application that supports NFTs.
 You will learn how to create a custom module that adds the feature to create and purchase NFTs in a blockchain application, and a custom plugin that provides additional NFT-related features.
+
+IMPORTANT: The example app for this tutorial was build without bootstrapping via Lisk Commander, therefore the project setup and file structure are different in this tutorial.
+If you wish to see an example how to easily bootstrap a blockchain application with the Lisk Commander, check out xref:{url_build_bapp}[how to bootstrap the Hello World application].
 
 TIP: For the *full code example* please see the {url_github_nft}[NFT app on Github^].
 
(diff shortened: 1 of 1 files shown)


docs/modules/ROOT/pages/tutorial/nft.adoc
```diff
@@ -121,11 +125,11 @@ The following functions are implemtented:
...
-** ownerAddress
-** minPurchaseMargin
-** name
+** id: The unique ID of the NFT.
+** value: The value of the NFT.
+** ownerAddress: The address of the owner of the NFT.
+** minPurchaseMargin: The minimum price increase, that should happen with every purchase of the NFT in percentage.
+** name: The NFT name.
 * `CHAIN_STATE_NFT_TOKENS`: The key under which the NFTs are saved in the database.
 * `createNFTToken()`: Creates a new NFT based on the provided `name`, `ownerAddress`, `nonce`, `value` and  `minPurchaseMargin`.
```
chris529:
```suggestion
* `createNFTToken()`: Creates a new NFT based on the provided `name`, `ownerAddress`, `nonce`, `value`, and  `minPurchaseMargin`.
```

docs/modules/ROOT/pages/tutorial/nft.adoc
```diff
@@ -520,6 +524,8 @@ Create a new file `purchase_nft_asset.js` inside the `transactions/` folder.
 
 Analog to the implementation of the `createNFT` asset, create the `purchaseNFT` asset by pasting the snippet below.
 
+NOTE: The validation of the asset inputs is done in the `apply()` function, because we need to access the database to validate the transaction inputs.
```
chris529:
```suggestion
NOTE: The validation of the asset inputs is done in the `apply()` function, as it is necessary to access the database in order to validate the transaction inputs.
```

chris529 (commented):
Looks good, just a couple of very minor points :

reyraa (approved):
Good job @Tschakki

PR #1503, opened 2022-07-11
Explain key concepts of the Lisk blockchain - Closes #1488

### LINKED ISSUE #1488: Add `Key concepts of Lisk SDK blockchain`
### Description

Create new page explaining the key concepts of the Lisk blockchain.

### ToC

- Give overview about the different domains + links to further explanations
  - Network
  - Consensus
  - State machine/application
- Transactions

```
The base transaction schema (without the params serialization) gives 117 bytes

Token transfer
- Minimum bytes: (117 + 36) bytes = 153 bytes
Delegate registration
- Minimum bytes: (117 + 191) bytes = 308 bytes
Delegate vote
- Minimum bytes: (117 + 30) bytes = 147 bytes 
Token unlock
- Minimum bytes: (117 + 0) bytes = 117 bytes 
Multisignature group registration
- Minimum bytes: (117 + 102) bytes = 219 bytes 
Delegate misbehavior report 
- Minimum bytes: (117 + 874) bytes = 991 bytes 
```

- Blocks

### Title

Key concepts of the Lisk blockchain

### Nav

`/understand-blockchain`

Tschakki:
Closes #1488


docs/modules/ROOT/pages/understand-blockchain/index.adoc
```diff
@@ -1,236 +1,502 @@
...
+:url_intro_consensus: intro/how-blockchain-works.adoc#consensus-mechanisms
+:url_understand_state_machine: understand-blockchain/state-machine.adoc
+:url_understand_consensus: understand-blockchain/consensus.adoc
+:url_understand_network: understand-blockchain/network.adoc
+:url_understand_state_machine_tree: understand-blockchain/state-machine.adoc#the-blockchain-state-as-sparse-merkle-tree
+:url_understand_sdk_commands: understand-blockchain/sdk/modules-assets.adoc#assets
+
+In the "Understand blockchain" category you learn the basic concepts of blockchain applications, including how they are structured and which topics and terms are important to know when developing within the Lisk ecosystem.
+
+This page explains the three domains of a blockchain, and introduces to blocks and transactions.
```
sergeyshemyakov: > introduces to blocks and transactions

I guess you wanted to say "introduces blocks and transactions" or "gives introduction to blocks and transactions"

docs/modules/ROOT/pages/understand-blockchain/index.adoc
```diff
@@ -1,236 +1,502 @@
...
+. xref:{url_understand_consensus}[Consensus domain]: Responsible for the replication of the same sequence of states among all nodes in the network.
+This is achieved by nodes in the network following a xref:{url_intro_consensus}[consensus protocol] and utilizing the state machine and network domains.
+. xref:{url_understand_network}[Network domain]: Responsible for the communication of the peer-to-peer network.
+
+== Blocks
+
+Blocks following the Lisk protocol have three main properties:
+
+. The <<block-header,block header>> includes properties relevant to the xref:{url_understand_consensus}[consensus] domain
+. The <<block-assets,block assets>> allow modules to include specific data into a block.
```
sergeyshemyakov: I think at this point the reader would not be aware of the modules and what they do. Maybe some further comments would be helpful, e.g. explain why assets are needed here (to support a piece of custom blockchain logic).

docs/modules/ROOT/pages/understand-blockchain/index.adoc
```diff
@@ -1,236 +1,502 @@
...
+====
+
+=== Block header
+
+Properties handled by the xref:{url_understand_consensus}[consensus] domain are added to the block header.
+
+.Block ID
+NOTE: The block ID is calculated by hashing the complete block header of a signed block.
+
+The block header consists of the following properties:
```
sergeyshemyakov: I would suggest leaving only the most important/ non-technical properties here like `timestamp`, `height`, `previousBlockID`, `generatorAddress`, `transactionRoot`, `signature`. Otherwise the section looks a bit overwhelming.

You could say: all properties are listed in the block header schema and refer to the LIP for more explanations on other properties.

docs/modules/ROOT/pages/understand-blockchain/index.adoc
```diff
@@ -1,236 +1,502 @@
...
+  }
+}
+----
+====
+
+=== Block generation
+
+image::understand-blockchain/block-generation.png["Block generation steps",500, align="center"]
+
+The full generation of a block is organized as follows.
```
sergeyshemyakov: I think it's important to explain and rationalize our block generating (and later block processing) architecture. I would comment that such block generation offers a lot of flexibility for custom logic without a virtual machine and makes all important verification steps explicit and obvious. Maybe you have some more arguments.

sergeyshemyakov (commented):
Just a couple of comments, otherwise looks good!

docs/modules/ROOT/pages/understand-blockchain/index.adoc
Tschakki: I tried to improve the explanation, let me know if you have some other suggestions!

docs/modules/ROOT/pages/understand-blockchain/index.adoc
Tschakki: Yes good point, I tried to add some rationale, let me know if i need to add/adjust something.

docs/modules/ROOT/pages/understand-blockchain/index.adoc
Tschakki: I kept the explanations of the less important properties, but hid them in a collapsible box

docs/modules/ROOT/pages/intro/what-is-blockchain.adoc
```diff
@@ -12,8 +12,12 @@ Chris Braithwaite <christopher.braithwaite@lightcurve.io> Mona Bärenfänger <mo
...
-The benefits of using this structure are that at its core, it is a shared and immutable ledger that enables transactions to be recorded, and assets to be tracked. Furthermore, with this technology it is possible to track any item of value, and also trade it on a blockchain network, which in turn offers numerous benefits, such as reducing the costs and risk for all parties, to name just a few.
+The term blockchain can be easily defined as a structure that has the ability to store transactional records.
+This method of storage is often referred to as a 'digital ledger'.
+It can be further broken down into 'blocks' which refer to how the data is stored, and then into 'chains' which pertains to the blocks of data being linked together in a permanent 'chain', resulting in a 'blockchain'.
+
+The benefits of using this structure are that at its core, it is a shared and immutable ledger that enables transactions to be recorded, and assets to be tracked.
+Furthermore, with this technology it is possible to track any item of value, and also trade it on a blockchain network, which in turn offers numerous benefits, such as reducing the costs and risk for all parties, to name just a few.
```
shuse2, line 20: I think below `What is a block` also needs slight update because now block has `assets` and `transactions` as payload

docs/modules/ROOT/pages/understand-blockchain/index.adoc
sergeyshemyakov: I think it's good now, since you don't mention modules explicitly anymore.

docs/modules/ROOT/pages/understand-blockchain/index.adoc
sergeyshemyakov: Looks very good now!

docs/modules/ROOT/pages/understand-blockchain/index.adoc
sergeyshemyakov: Nice explanation!

sergeyshemyakov (approved):
Now it looks very nice!

docs/modules/ROOT/pages/understand-blockchain/index.adoc
```diff
@@ -1,236 +1,520 @@
...
+                "length": 64
+            },
+            "fieldNumber": 7
+        }
+    }
+}
+----
+====
+
+=== Default transaction types
```
shuse2, line 493: Terminology wise, i think we should say `commands that modules SDK provides`?
but I think it's better not to add this set here because this is explaining the concept, not specific feature

docs/modules/ROOT/pages/understand-blockchain/sdk/index.adoc
```diff
@@ -0,0 +1,236 @@
+= Introduction to the Lisk SDK
```
shuse2, line 1: is this page scope of this PR? maybe this page is duplicated or moved, but it contains outdated information

docs/modules/ROOT/pages/understand-blockchain/index.adoc
```diff
@@ -1,236 +1,520 @@
-= Key concepts of the Lisk SDK blockchain
-Mona Bärenfänger <mona@lightcurve.io>
+= Key concepts of the Lisk blockchain
```
shuse2, line 1: The images refers to `state machine domain`, but I think in LIP, it's updated to say `application domain` because in our SDK architecture, how we divided is not only about state machine. (ie: the `state machine` domain above contains concepts which are not `state machine`)

shuse2 (changes requested):

docs/modules/ROOT/pages/understand-blockchain/sdk/index.adoc
Tschakki: No, it is not scope of this PR. Only the page "key concepts of Lisk blockchain" is the scope

docs/modules/ROOT/pages/understand-blockchain/index.adoc
Tschakki: As far as it was communicated, we always used `state machine`, and also in the docs site structure we planned it like this. If we really want to change this now, it should be updated there first, as this is the main source for creating the new docs structure, including names of pages.

docs/modules/ROOT/pages/understand-blockchain/index.adoc
Tschakki: I think the table is super helpful and fits on the page, as it explains transactions to the reader. It is helpful to see an overview of existing examples while reading about the concept. Also, I don't know where it would fit better.

Tschakki: And I would talk here about transaction types and then introducing commands in relation to transaction types. Otherwise it will be confusing for the reader imho. I think the connection between commands and transaction is explained sufficient on this page.

docs/modules/ROOT/pages/intro/what-is-blockchain.adoc
Tschakki: This is not scope of this PR

Tschakki: I think in `intro` section we talk about what is blockchain in general, not focussing on Lisks solution. And `understand-blockchain` is then the category, where we focus on the Lisk solution. In that context, I think it is fine to not mention block assets on this page, as I believe this property is pretty Lisk specific?

docs/modules/ROOT/pages/understand-blockchain/sdk/index.adoc
Tschakki: It was the previous `understand-blockchain/index.adoc` page, I moved it for now but it will be updated in #1509

docs/modules/ROOT/pages/intro/what-is-blockchain.adoc
shuse2: ah ok, yes, assets is very lisk specific 👍  make sense

docs/modules/ROOT/pages/understand-blockchain/index.adoc
shuse2: I think the concrete commands will fit better on each module explanation?
Interoperability module is actually MainchainInteroperability/SidechainInteroperability module (2 different modules), and we will have NFT/PoA and more in the future, so it will be quite a long list.

docs/modules/ROOT/pages/understand-blockchain/index.adoc
shuse2: I think explaining `state machine` is fine.
The difference is when we split in to 3 parts, in the LIP55, we said we split in application/consensus/network domain, and state machine is part of the `application` domain.

It is because not all `module` logic are strictly speaking `state machine`. For example, `insertAsset` or `verifyTransaction` etc hooks are not mutating the state.

Tschakki: Ok, I have now updated the images and descriptions to mention the `application` domain instead of `state machine`

docs/modules/ROOT/pages/understand-blockchain/index.adoc
Tschakki: ok, I commented the table out, in case I can reuse parts of it somewhere else :)

shuse2 (approved):

docs/modules/ROOT/pages/intro/what-is-blockchain.adoc
```diff
@@ -12,8 +12,12 @@ Chris Braithwaite <christopher.braithwaite@lightcurve.io> Mona Bärenfänger <mo
...
 
 == What is a block?
```
gkoumout, line 22: in the last paragraph of this section, I guess the last phrase ", (each separate node) " should just be removed

docs/modules/ROOT/pages/understand-blockchain/index.adoc
```diff
@@ -1,236 +1,516 @@
...
+. xref:{url_understand_consensus}[Consensus domain]: Responsible for the replication of the same sequence of states among all nodes in the network.
+This is achieved by nodes in the network following a xref:{url_intro_consensus}[consensus protocol] and utilizing the application and network domains.
+. xref:{url_understand_network}[Network domain]: Responsible for the communication of the peer-to-peer network.
+
+== Blocks
+
+Blocks following the Lisk protocol have three main properties:
+
+. The <<block-header,block header>> includes properties relevant to the xref:{url_understand_consensus}[consensus] domain.
+. The <<block-assets,block assets>> are an array of objects containing data injected by the blockchain application during the block creation.
```
gkoumout: Perhaps rephrase: "block assets are an array...".

Example: `` Block assets: An array of objects ..."

docs/modules/ROOT/pages/understand-blockchain/index.adoc
```diff
@@ -1,236 +1,516 @@
...
+    },
+    "data": {
+      "dataType": "bytes",
+      "fieldNumber": 2
+    }
+  }
+}
+----
+====
+
```
gkoumout, line 271: Perhaps add the transactions section here; this would make it more consistent since the block consist of: header, assets, transactions. 

Now the order is a bit arbitrary by defining just header and asset, then proceed to block generation and then going back to transactions. 

If the reason for the current structure is that transactions section is kind of more important and special, we can just add here a small comment that transactions will be discussed in a later section and add a link to it.

docs/modules/ROOT/pages/understand-blockchain/index.adoc
gkoumout: Actually I think that it is a good idea to have some table like that here. Somehow the first-time reader gets a sense of what is module is doing by viewing intuitive transaction names here (e.g,. register delegate in DPos, token transfer in Token etc.)

 I can see the point that it gets too long, but I would not view this as an issue. If you insist that a long table is bad, as an intermediate solution we could have an example like: 
 
Token Module includes the following transactions --> table
 DPos Module includes the following transactions  --> table
 
 Similarly all modules define transactions, see documentation/LIP, etc.

gkoumout (changes requested):
Looks good in general! 

Some conctrete comments:

- In the "what is blockchain" file, the section "How does a blockchain transaction work" seems confusing in steps 3 and 4, since the term block (and hashing) is first mentioned in point 3 and then in point 4 the block creation step is described. Perhaps say more explicilty in point 3 that a transaction is included in the block. 

- In the "Key concepts of Lisk Blockchain file", in a later stage a link to the transaction schema LIP should be added (currently a PR is opened, but not merged yet)

docs/modules/ROOT/pages/understand-blockchain/index.adoc
Tschakki: Yes valid point, I now moved the transaction section up, let me know if it's ok now!

gkoumout (approved):
No complaints! All my comments addressed  ;-)

PR #1586, opened 2022-07-25
Add new page about conensus protocols - Closes #1492

### LINKED ISSUE #1492: Add page about block generation & consensus mechanisms
### Description

Create new page explaining what are consensus algorithms.
### ToC

- Introduction to consensus algorithms: what is it, why do we need it
- Block generation & finalization
- Different types of Consensus algorithms
- Certificate generation

### Title

Consensus algorithm

### Nav

`/understand-blockchain/consensus`

Tschakki:
Closes #1492


mitsuaki-u (approved):

docs/modules/ROOT/pages/understand-blockchain/consensus/index.adoc
```diff
@@ -0,0 +1,46 @@
+= Consensus protocols
+:toc:
+:fn_fork: footnote:eventroot[Fork: A blockchain has forked, if two or more conflicting states(i.e. chain of blocks) of the blockchain exist in the network, and the network is unable to decide which is the correct state.]
+:fn_crosschain: footnote:crosschain[Cross-chain: Means that two or more different blockchains are involved.]
+:fn_validators: footnote:validators[Validators: Accounts eligible to add new blocks to the chain.]
```
janhack: Maybe we could avoid footnotes here and rather add these things to the glossary? The footnotes make things difficult to read and also the layout does not look so nice.

docs/modules/ROOT/pages/understand-blockchain/consensus/index.adoc
```diff
@@ -0,0 +1,46 @@
...
+
+// Project URLs
+:url_understand_dpos: understand-blockchain/consensus/dpos.adoc
+:url_understand_poa: understand-blockchain/consensus/poa.adoc
+:url_understand_bft: understand-blockchain/bft.adoc
+:url_understand_interop: understand-blockchain/interoperability.adoc
+:url_intro_consensus: intro/how-blockchain-works.adoc#consensus-mechanisms
+
+A consensus protocol is the protocol for validators{fn_validators} to **agree on one chain of blocks**.
```
janhack: This is one aspect of consensus protocols, but does not cover everything, see first section in https://lisk.com/documentation/understand-blockchain/lisk-protocol/consensus-algorithm.html.

docs/modules/ROOT/pages/understand-blockchain/consensus/index.adoc
```diff
@@ -0,0 +1,46 @@
...
+TIP: There are multiple available protocols for the consensus domain `Block proposal` that you can choose from, to fit the specific use case.
+
+[cols="1a,1,2a"]
+|===
+|Consensus domain|Consensus protocol|Responsible for
+
+|Block proposal
+|xref:{url_understand_dpos}[Lisk DPoS] or xref:{url_understand_poa}[Lisk PoA]
+|* **Choosing validators**: Deciding who can be validator, i.e. who is allowed to add new blocks to the blockchain.
+* **Block frequency**: Defining when a block is added to the chain.
```
janhack: Do you mean here, for instance, that a block is added every 10 seconds? I don't think the block time is defined in the PoA or DPoS module at the moment.

docs/modules/ROOT/pages/understand-blockchain/consensus/index.adoc
```diff
@@ -0,0 +1,46 @@
...
+
+|Agreement on one chain
+|xref:{url_understand_bft}[Lisk BFT]
+|* **Fork{fn_fork} prevention**: Determining how validators choose the chain to which they add a new block.
```
janhack: I would rather call this "fork choice rule" as it does not prevent forks, but rather helps to resolve them.

docs/modules/ROOT/pages/understand-blockchain/consensus/index.adoc
```diff
@@ -0,0 +1,46 @@
...
+* **Chain finalization**: Determining when a block (and all included transactions) is irrevocably deemed part of the blockchain.
```
janhack: I would rather call it "block finalization" as generally not the whole chain is finalized (especially not the last one just added), but rather one block and all its ancestors.

docs/modules/ROOT/pages/understand-blockchain/consensus/index.adoc
```diff
@@ -0,0 +1,46 @@
...
+
+|Cross-chain{fn_crosschain} agreement
```
janhack: I would rather put "certificate generation" here. There is not really a cross-chain agreement protocol, but rather certificate generation (in the sending blockchain) and certificate validation (in the receiving blockchain).

docs/modules/ROOT/pages/understand-blockchain/consensus/index.adoc
```diff
@@ -0,0 +1,46 @@
...
+|xref:{url_understand_interop}[Lisk Interoperability]
+|* **Verifying cross-chain communication**: Attesting the correctness of cross-chain messages with certificates.
```
janhack: I would rather focus on the fact that certificates are automatically generated for finalized blocks. This happens now in the engine and is closely connected to the Lisk BFT protocol. You could mention the 3 steps in the certificate generation process (see blog post https://lisk.com/blog/research/certificate-generation-and-consensus) if you want to go into detail.

We could also mention that the certificates generation process is designed in a way to ensure that the generated certificates satisfy the chain of trust property.

docs/modules/ROOT/pages/understand-blockchain/consensus/index.adoc
```diff
@@ -0,0 +1,46 @@
...
+* **Chain of trust for validators **: Attesting updates about the current validator set of a chain in certificates.
```
janhack: If we want to say something about chain of trust, I would suggest something like:

> Generated certificates satisfy chain of trust property, i.e., they can be sequentially submitted to another blockchain via cross-chain update transactions.

janhack (changes requested):
Overall, the page looks goods. I left a some comments and suggestions to improve the clarity in a few places.

docs/modules/ROOT/pages/understand-blockchain/consensus/index.adoc
Tschakki: Updated it to include more aspects, and structured it in a bullet list for better overview

docs/modules/ROOT/pages/understand-blockchain/consensus/index.adoc
Tschakki: ah ok, then I will remove it here

docs/modules/ROOT/pages/understand-blockchain/consensus/index.adoc
Tschakki: Hm, but don't the certificates in a sense enable cross-chain agreement about a certain state of the blockchain across chains? My thought of calling it `cross chain agreement` was that it explains the purpose in a more general way than `certificate generation`. Because for the later you would need to know what are certificates and what is their purpose in the application.

docs/modules/ROOT/pages/understand-blockchain/consensus/index.adoc
Tschakki: I added now additional information. But please take into account we only talk about certificates in a very short and general manner here, the details are then mentioned on the respective explanation of the consensus protocol, here as part of the Lisk interoperability solution.

docs/modules/ROOT/pages/understand-blockchain/consensus/index.adoc
```diff
@@ -0,0 +1,51 @@
...
+
+* *who* is allowed to add a new block to the blockchain,
+* how delegates choose the *chain* to which they add a new block and
+* when a block and all included transactions are **final**, i.e., irrevocably deemed part of the blockchain.
+
+TIP: See also xref:{url_intro_consensus}[Intro > How blockchain works > Consensus mechanisms].
+
+== Consensus protocols in Lisk
+
+Blockchain applications built with the Lisk SDK achieve consensus in all important domains via dedicated modules, which each cover a specific aspect of the overall consensus of a blockchain.
```
janhack: Note that "Lisk-BFT" is no moved to the engine and will not be a module any more.

janhack:
```suggestion
In the Lisk SDK, the different aspects of the consensus protocol are handled by different sub-protocols as outlined in the table below. 
```

docs/modules/ROOT/pages/understand-blockchain/consensus/index.adoc
```diff
@@ -0,0 +1,51 @@
...
+A consensus protocol is the protocol for xref:{url_glossary_validator}[validators] to **agree on one chain of blocks**.
+In particular, the consensus protocol defines,
+
+* *who* is allowed to add a new block to the blockchain,
+* how delegates choose the *chain* to which they add a new block and
+* when a block and all included transactions are **final**, i.e., irrevocably deemed part of the blockchain.
+
+TIP: See also xref:{url_intro_consensus}[Intro > How blockchain works > Consensus mechanisms].
+
+== Consensus protocols in Lisk
```
janhack:
```suggestion
== Consensus protocol in Lisk
```

janhack: I would say there is one consensus protocol which is organized into different sub-protocols.

docs/modules/ROOT/pages/understand-blockchain/consensus/index.adoc
janhack: Cross-chain agreement suggests that different blockchains communicate directly with each other (however, this only happens if there is a relayer) and there could be information flow into both direction. However, in the cross-chain certification approach, one blockchain just informs other blockchains about their state with the certificate signatures to attest for the correctness. So it is very much only a one-direction information flow that only happens if there is a relayer.

docs/modules/ROOT/pages/understand-blockchain/consensus/index.adoc
```diff
@@ -0,0 +1,51 @@
...
+|xref:{url_understand_bft}[Lisk BFT]
+|* **xref:{url_glossary_fork}[Fork] choice rule**: Determining how validators choose the chain to which they add a new block.
+* **Block finalization**: Determining when a block (and all included transactions) is irrevocably deemed part of the blockchain.
+
+|xref:{url_glossary_cchain}[Cross-chain] agreement
+|xref:{url_understand_interop}[Lisk Interoperability]
+|* **Chain of trust for validators **: Attesting updates about the current validator set of a chain in certificates.
+* **Verifying cross-chain communication**: Attesting the correctness of cross-chain messages with certificates.
+Certificates are automatically generated for finalized blocks.
+Generated certificates satisfy chain of trust property, i.e., they can be sequentially submitted to another blockchain via cross-chain update transactions.
```
janhack: As mentioned before, I don't think the cross-chain communication belongs to the consensus protocol. Instead, I would suggest to remove the bullet point "Verifying cross-chain communication" and rename "[Cross-chain] agreement" to "Certificate generation"

docs/modules/ROOT/pages/understand-blockchain/consensus/index.adoc
janhack: You could have one bullet point here:
> - "**Certificate generation for interoperability**: Certificates are generated for finalized blocks. Certificates are crucial for submitting cross-chain updates to other blockchains as they attest outgoing cross-chain messages and changes of validators."

janhack (changes requested):
I added a few more comments and suggestions.

docs/modules/ROOT/pages/understand-blockchain/consensus/index.adoc
```diff
@@ -0,0 +1,50 @@
...
+
+== Consensus protocol in Lisk
+
+In the Lisk SDK, the different aspects of the consensus protocol are handled by different sub-protocols as outlined in the table below. 
+
+TIP: There are multiple available protocols for the consensus domain `Block proposal` that you can choose from, to fit the specific use case.
+
+[cols="1a,1,2a"]
+|===
+|Consensus domain|Consensus protocol|Responsible for
```
janhack:
```suggestion
|Domain|Protocol|Responsible for
```

docs/modules/ROOT/pages/understand-blockchain/consensus/index.adoc
```diff
@@ -0,0 +1,50 @@
...
+|* **Choosing validators**: Deciding who can be validator, i.e. who is allowed to add new blocks to the blockchain.
+
+|Agreement on one chain
+|xref:{url_understand_bft}[Lisk BFT]
+|* **xref:{url_glossary_fork}[Fork] choice rule**: Determining how validators choose the chain to which they add a new block.
+* **Block finalization**: Determining when a block (and all included transactions) is irrevocably deemed part of the blockchain.
+
+|Certificate generation
+|xref:{url_understand_interop}[Lisk Interoperability]
+|* **Chain of trust for validators **: Attesting updates about the current validator set of a chain in certificates.
```
janhack: I would remove this part as the chain of trust property of the generated certificates is not that important here in my opinion and it does not really fit under "Responsible for".

docs/modules/ROOT/pages/understand-blockchain/consensus/index.adoc
```diff
@@ -0,0 +1,50 @@
...
+|Block proposal
+|xref:{url_understand_dpos}[Lisk DPoS] or xref:{url_understand_poa}[Lisk PoA]
+|* **Choosing validators**: Deciding who can be validator, i.e. who is allowed to add new blocks to the blockchain.
+
+|Agreement on one chain
+|xref:{url_understand_bft}[Lisk BFT]
+|* **xref:{url_glossary_fork}[Fork] choice rule**: Determining how validators choose the chain to which they add a new block.
+* **Block finalization**: Determining when a block (and all included transactions) is irrevocably deemed part of the blockchain.
+
+|Certificate generation
```
janhack: Maybe the domain could be "Certificates" and the protocol "Certificate generation protocol" or "Certificate generation mechanism". Note that this protocol is defined in [LIP 61](https://github.com/LiskHQ/lips/blob/main/proposals/lip-0061.md) and rather an add-on to the consensus, then the Lisk Interoperability protocol (that is rather what is defined in the Interoperability module).

janhack (changes requested):
I still have some minor comments.

docs/modules/ROOT/pages/understand-blockchain/consensus/index.adoc
Tschakki: Hm, but where in the current structure are we explaining certificates? I was imaging it is done in one of the pages explaining the Lisk interoperability solution. If it doesn't fit there, maybe we should think of adding an additional page, just talking about certificates, and linking to it here.

janhack (approved):
Looks all good.

PR #1598, opened 2022-08-08
Updates Create User Interface Guide - Closes 1589

### LINKED ISSUE #1589: Update `Create User Interface Guide` to address possible errors
### Problem
Whilst going through the steps of the [Create User Interface Guide](https://lisk.com/documentation/integrate-blockchain/create-user-interface.html) the reader is asked to fill out the contents of the `Faucet.js` file. Inside the code, is a line importing an `accounts.json` file like this: `import accounts from '../accounts.json';`.

This code should import a file with a list of all the accounts created. However, the guide doesn't inform the user about this file. As a result, an error occurs for readers who had previously generated a genesis block as suggested in the [How to create a new genesis block guide](https://lisk.com/documentation/build-blockchain/create-genesis-block.html). The users who are using the code from earlier examples will face this issue. Also since the guide doesn't mention the `accounts.json` file, readers trying out the instructions will miss the addition of such a file into the `hello_frontend` application

### Solution
The solution is to direct the reader to import the `accounts.json` file from their `hello_app` (blockchain project). This will help the reader in two cases. 

- If the reader had already created a new genesis block, they will have to run their application with the new set of accounts and a new `custom_config`.
  - By linking to the `accounts.json` from the blockchain project, both the `hello_app` (blockchain) and `hello_frontend`will refer to the same `accounts.json` file.
  - Even if the reader doesn't create a new genesis block, pointing to the same file in both projects will work and any inconsistency in this regard will be avoided.

- After telling the reader about linking the said file, we should remove the following code snippet from the `Faucet.js`:
```
const accounts = {
  "genesis": {
    "passphrase": "peanut hundred pen hawk invite exclude brain chunk gadget wait wrong ready"
  }
};
```
- After that, we should also update the `Faucet.js: LN 36` which is `}, accounts.genesisDelegate.passphrase);` to something that allows the code to read the first account in the `accounts.json` file.

Affected branch: `main/development`

TalhaMaliktz:
Closes #1589

### DIFF
--- a/docs/modules/ROOT/pages/integrate-blockchain/create-user-interface.adoc
+++ b/docs/modules/ROOT/pages/integrate-blockchain/create-user-interface.adoc
@@ -64,7 +64,7 @@ mv ~/<YOUR_DIRECTORY>/hello_app ~/<YOUR_DIRECTORY>/HelloApplication
 
 == Create a new React app
 
-Bootstrap the React app with the following command:
+While being in the `HelloApplication` folder, bootstrap the React app with the following command:
 
 [source,bash]
 ----
@@ -317,6 +317,13 @@ const Faucet = () => {
 
 export default Faucet;
 ----
+Both frontend and blockchain applications should have the same `accounts.json`. So, copy `accounts.json` from `hello_app` and paste it into `hello_frontend`. 
+
+[source,bash]
+----
+cp hello_app/config/default/accounts.json hello_frontend/src/
+----
+
 
 === Send Transfer transaction
 
--- a/docs/modules/ROOT/pages/integrate-blockchain/create-user-interface.adoc
+++ b/docs/modules/ROOT/pages/integrate-blockchain/create-user-interface.adoc
@@ -200,7 +200,7 @@ export default NewAccount;
 
 The faucet is a component that allows accounts to receive tokens from the genesis account, which holds the majority of initial tokens at the start of the Devnet.
 
-In a new file `api.js`, the `apiClient` from package `lisk-client` provides an interface for the faucet and other React components to connect to the blockchain application via a websocket on port 8080.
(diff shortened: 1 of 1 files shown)


docs/modules/ROOT/pages/integrate-blockchain/create-user-interface.adoc
```diff
@@ -44,6 +44,23 @@ To use this guide, it is assumed that the following criteria have been met:
...
+----
+mv ~/<YOUR_DIRECTORY>/hello_app ~/<YOUR_DIRECTORY>/HelloApplication
+----
+
+----
+├── HelloApplication/
+│   └── hello_app/
+----
 
 == Create a new React app
```
Tschakki, line 66: In the next line, it is also good to mention the folder they should be in when running the command. Something like

`While being in the `HelloApplication` folder, bootstrap the React app with the following command:`

docs/modules/ROOT/pages/integrate-blockchain/create-user-interface.adoc
```diff
@@ -219,13 +244,8 @@ import React, { useState } from 'react';
...
 import * as api from '../api.js';
-
-// The passphrase for a genesis account of the local Devnet.
-const accounts = {
-  "genesis": {
-    "passphrase": "peanut hundred pen hawk invite exclude brain chunk gadget wait wrong ready"
-  }
-};
+//List of accounts, the list should be same as the hello_app
+import accounts from '../accounts.json';
```
Tschakki, line 248: A step describing how to copy the `accounts.json ` file might be helpful here

Tschakki (changes requested):

Tschakki (approved):
LGTM

PR #1629, opened 2022-09-19
wrong field number in "NFT Tutorial"

HadiEsna:
wrong field number in accountSchema


Tschakki:
Hi, thank you for the fix, but unfortunately it is created on the wrong base branch; please use `main` instead, thanks!

PR #1630, opened 2022-09-20
fixed wrong type in config.json

Bullrich:
The page [How to configure a blockchain application](https://lisk.com/documentation/build-blockchain/configure-app.html) indicated the user to copy a configuration, but the type for `.genesisConfig.baseFees.0.baseFee` is a string and in the current example it is set as a number.

If you copy and paste the current implementation it will throw an error:
```
Error: Lisk validator found 1 error[s]:
Property '.genesisConfig.baseFees.0.baseFee' should be of type 'string'
```

This should fix that issue. I believe it will also resolve #1521 as I didn't find any other problem in that page.

PR #1633, opened 2022-09-21
 fixed wrong type in config.json #1630 

### LINKED ISSUE #1521: Update `How to configure a blockchain application`
### Description

* Update `How to configure a blockchain application` to use Lisk SDK v6 config schema
* Update terminologies to Lisk SDK v6
* Add module / plugin config explanation

### Additional Information
Original: https://lisk.com/documentation/build-blockchain/configure-app.html
SDK reference: https://github.com/LiskHQ/lisk-sdk/blob/development/framework/src/schema/application_config_schema.ts

- Use hello world app configuration as example

Bullrich:
The page [How to configure a blockchain application](https://lisk.com/documentation/build-blockchain/configure-app.html) indicated the user to copy a configuration, but the type for `.genesisConfig.baseFees.0.baseFee` is a string and in the current example it is set as a number.

If you copy and paste the current implementation it will throw an error:
```
Error: Lisk validator found 1 error[s]:
Property '.genesisConfig.baseFees.0.baseFee' should be of type 'string'
```

This should fix that issue.

### DIFF
--- a/docs/modules/ROOT/pages/build-blockchain/configure-app.adoc
+++ b/docs/modules/ROOT/pages/build-blockchain/configure-app.adoc
@@ -159,7 +159,7 @@ TIP: To see a full list of all constants and their predefined values, please see
             {
                 "moduleID": 2,
                 "assetID": 0,
-                "baseFee": 1000000
+                "baseFee": "1000000"
             }
         ],
         "activeDelegates": 31,


Tschakki (approved):
Thanks for the fix! 👍 

But #1521 is related to the next version of the SDK, and is not closed by this PR.

Bullrich:
> Thanks for the fix! 👍
> 
> But #1521 is related to the next version of the SDK, and is not closed by this PR.

Ok, I have updated the PR body to remove that text

PR #1683, opened 2022-12-27
Update Module ID related content according to v6 - Closes 1680

### LINKED ISSUE #1680: Module ID has been depreciated
### Description

1. The following text is invalid as the Module ID has been replaced with the Module name in SDK v6.

> Module ID: The ID of the module, identifying the key-value map. We will reserve all module IDs starting with a 0 in their binary representation to modules that are shipped with the SDK by default. The IDs of modules developed as part of each blockchain application will start with a 1. The module ID has a fixed length of 4 bytes.

2. The text in the https://liskhq.github.io/lisk-docs/understand-blockchain/blocks-txs.html page also has the` Module ID` and `Command ID` in text and in snippets as well which should be updated.

### Affected branch / URL
[URL of the page affected by the bug. 1](https://liskhq.github.io/lisk-docs/understand-blockchain/state-machine.html#the-blockchain-state-as-sparse-merkle-tree)

[URL of the page affected by the bug. 2](https://liskhq.github.io/lisk-docs/understand-blockchain/blocks-txs.html)

`development`

TalhaMaliktz:
Closes #1680


Tschakki (changes requested):
I just checked, there is also content in the Blocks & transaction page that needs to be updated, because it still mentions module and command IDs: https://liskhq.github.io/lisk-docs/understand-blockchain/blocks-txs.html
Here block asset schema and transaction schema need update as well

Tschakki (approved):
LGTM 👍

PR #1693, opened 2023-01-09
Fix links in SDK docs

### LINKED ISSUE #1692: Update and fix links to the root component
### Description
All pages need to be checked for links to the root component `ROOT::`.

If a link exists, it should be check if it needs to be replaced with `beta@ROOT::`

#### Affected branch / URL
`docs-sdk`, `docs-core`

### Additional info
This update is needed, because we decided to include two versions of the docs root component, as long as SDK v6 is still in beta

Tschakki:
Partially closes #1692


TalhaMaliktz (approved):
LGTM 😊

PR #1750, opened 2023-03-21
Add setup relayer node guide - Closes 1534

### LINKED ISSUE #1534: Add `Setup relayer node`
### Description
- Explain how to install and run relayer node for 2 chains
  - Explain how to run a node with chain connector plugin for sidechain to mainchain communication
  - Explain how to run a node with chain connector plugin for mainchain to sidechain communication

### Additional Information
- This information depends on https://github.com/LiskHQ/lisk-sdk/issues/6918

TalhaMaliktz:
Closes #1534


docs/modules/ROOT/pages/run-blockchain/setup-relayer.adoc
```diff
@@ -0,0 +1,331 @@
...
+However, in case it is required to install it manually, run the following command in the root folder of the client.
+
+[source,bash]
+----
+npm install @liskhq/lisk-framework-chain-connector-plugin
+----
+
+== Configuring the Chain Connector plugin
+Once the plugin is installed, various configurations are available for the node operator to set.
+With these configurations, the details such as the receiver’s chainID, the receiving chain’s path, CCU frequency, private key, etc. can be set.
```
ishantiw:
```suggestion
With these configurations, the details such as the receiver’s chainID, the receiving chain’s path, CCU frequency, encrypted private key, etc. can be set.
```

docs/modules/ROOT/pages/run-blockchain/setup-relayer.adoc
```diff
@@ -0,0 +1,331 @@
...
+
+// // Project URLS
+:url_guides_reg_sidechain: build-blockchain/register-sidechain.adoc#how-to-register-a-sidechain-on-the-mainchain
+:url_guides_reg_mainchain: build-blockchain/register-sidechain.adoc#how-to-register-the-mainchain-on-the-sidechain
+
+====
+On this page, you'll learn how to:
+
+* [x] How to set up a relayer node on a sidechain?
+* [x] How to set up a relayer node on a mainchain?
```
Tschakki:
```suggestion
On this page, you'll learn:

* [x] How to set up a relayer node on a sidechain
* [x] How to set up a relayer node on a mainchain
```

docs/modules/ROOT/pages/run-blockchain/setup-relayer.adoc
```diff
@@ -0,0 +1,331 @@
...
+The subsequent sections describe the steps to set up a relayer node.
+
+
+.Prerequisites
+[NOTE]
+====
+To use this guide, it is assumed that the following criteria have been met:
+
+. A sidechain has been registered on the mainchain, as explained in the guide: xref:{url_guides_reg_sidechain}[How to register a sidechain on the mainchain?]
+. The mainchain has been registered on the sidechain, as described in the guide: xref:{url_guides_reg_mainchain}[How to register the mainchain on the sidechain?]
```
Tschakki: In the context of the guide, i don't think this is accurate.
As the guide is targeting node operators, the requirement is, that they have set up a node connected to a blockchain, and have API access to a node on the receiving chain.

docs/modules/ROOT/pages/run-blockchain/setup-relayer.adoc
```diff
@@ -0,0 +1,331 @@
...
+--
+[source,bash]
+----
+lisk-core console
+Entering Lisk REPL: type `Ctrl+C` or `.exit` to exit
+----
+--
+=====
+
+. Call the `encrypt` function to create an encrypted private key and pass the `generatorPrivateKey` of your account to it.
```
Tschakki: Please describe quickly, how they can obtain the `generatorPrivateKey` in case it is not clear to the user

docs/modules/ROOT/pages/run-blockchain/setup-relayer.adoc
```diff
@@ -0,0 +1,331 @@
...
+--
++
+Mainchain::
++
+--
+[source,bash]
+----
+lisk-core> lisk.cryptography.encrypt.stringifyEncryptedMessage(await lisk.cryptography.encrypt.encryptMessageWithPassword('0d7501d3d5c9accaefb3c0b6a569473b59391ae406f6324f98fa6dd70e119368a6454f898d3b82c41b158206c72ecfe917a1071c8084b496a0c5867afc10830b', 'lisk'))
+
+'kdf=argon2id&cipher=aes-256-gcm&version=1&ciphertext=57db80457db93a1abeceee5c6f951ca04579c447a06f45cf5e8b5398e207a26da53a6b191a02c479ede455950eacb48f32d6609f2cd4b5a1a51e895b210b587ef046e6c3151ef2212efd0808b45328742d09a279e7d667f1670ff02a2fd5c91f4afd0a08efb8e6e90b0b11e93b15da8daaeea543a0ff54f3dd51c66cac3b04c6&mac=7822258b12e0c787f5bd622c562914438a9d74ca1e11e11b840f3001a678b04f&salt=d4d051a123326ad2b82c022603e790b6&iv=0bb9e76cd5163d6c5af9d89d&tag=fbcdb355b5135d48df948841de5fcdf5&iterations=1&parallelism=4&memorySize=2024'
```
Tschakki: Please separate command and response in the code snippets for better overview

docs/modules/ROOT/pages/run-blockchain/setup-relayer.adoc
```diff
@@ -0,0 +1,331 @@
...
+----
+
+The client must be rebuilt to entertain the changes to the code.
+
+[source,bash]
+----
+npm run build
+----
+
+Alternatively, the plugin can also be enabled using the `--enable-chain-connector-plugin` flag whilst starting the sidechain client, without updating the *app.ts* file.
```
Tschakki: Is this the alternative or the recommended way to enable the plugin?

Tschakki (changes requested):

docs/modules/ROOT/pages/run-blockchain/setup-relayer.adoc
TalhaMaliktz: It is an alternative way, there is no recommended way TBH. A user can go with any given method to enable the plugin as they like. Both are correct ways to do the same thing.

docs/modules/ROOT/pages/run-blockchain/setup-relayer.adoc
TalhaMaliktz:
```suggestion
On this page, you'll learn:

* [x] How to set up a relayer node on a sidechain?
* [x] How to set up a relayer node on a mainchain?
```

docs/modules/ROOT/pages/run-blockchain/setup-relayer.adoc
Tschakki: I see. Maybe we can just change order of the two options, as this one looks more simple, but without recommending one option directly?

docs/modules/ROOT/pages/run-blockchain/setup-relayer.adoc
TalhaMaliktz: Since registration is the last step for a node to be connected to the blockchain, I think the pre-requisite is accurate as it tells exactly what needs to be done for a node to be connected to the blockchain, and since a node operator or a validator will most probably be registering the chains I think it suffices to say that registration is necessary. 
If blockchain registers, it will automatically mean that the node of the blockchain are connected to the network and API access to nodes will be possible on receiving chain.

docs/modules/ROOT/pages/run-blockchain/setup-relayer.adoc
TalhaMaliktz: I have made the adjustment as suggested.

docs/modules/ROOT/pages/run-blockchain/setup-relayer.adoc
Tschakki: The sidechain/mainchain registration is not something every node operator needs to perform in order to connect to the network. It is only required one time to do the registration process.

For the guide you are assuming that the reader has setup a node connected to the sending chain, and has access to node API of the receiving chain. Probably it should be recommended, that the user who is setting up a relayer also manages the node on the receiving chain, to ensure availability?

So in this case we can mention the following requirements:

1. Managing a node connected to the mainchain (link to Lisk Core)
2. Managing a node connected to the sidechain
3.  The sidechain should be registered to the mainchain according to the guide "How to register a sidechain"(link) and the status of the sidechain account should be `registered` or `active`.

docs/modules/ROOT/pages/run-blockchain/setup-relayer.adoc
```diff
@@ -0,0 +1,348 @@
...
+On a sidechain, the plugin can be enabled using the `--enable-chain-connector-plugin` flag whilst starting the sidechain client.
+
+[source,bash]
+----
+./bin/run --enable-chain-connector-plugin --overwrite-config
+----
+
+Alternatively, the *app.ts* file of the client can be updated to have the following options:
+
+.sidechain_client/src/app/app.ts
```
Tschakki: The plugin should be registered in the `plugins.ts` file instead off `app.ts`, to use the current file structure.

docs/modules/ROOT/pages/run-blockchain/setup-relayer.adoc
```diff
@@ -0,0 +1,348 @@
...
+--
+[source,bash]
+----
+2023-03-17T14:42:30.426Z INFO XYZ.local plugin_chainConnector 96899 No valid CCU can be generated for the height: 58
+----
+--
+=====
+Since we just set up a relayer node and haven't sent a CCU/CCM, the aforementioned log messages are expected.
+
+The messages suggest that the blockchain doesn't have any finalized block height for which we can create a certificate and there are no pending CCUs/CCMs to send across the chain.
```
Tschakki: Isn't a first CCM created during the registration process on both chains? So that it can be expected that there is already 1 CCM present directly after registration? And if a node operator enables the chain connector plugin after the registration, how likely is it, that there are no CCMs to process when the plugin is enabled the first time? In case the default scenario is, that there are CCMs to be processed, so that it doesn't log the described error, it should explain the success message instead.

Tschakki (changes requested):

docs/modules/ROOT/pages/run-blockchain/setup-relayer.adoc
```diff
@@ -0,0 +1,348 @@
...
+----
+--
+=====
+
+
+. The `encryptMessageWithPassword` function will return an encrypted key, which should be added to the config of the blockchain.
+
+=== Prepare config for chain connector plugin
+There's a slight difference in the `chainConnector` config of a sidechain and a mainchain. 
+The mainchain expects the `chainID` of the receiving chain along with the mandatory fields.
```
Tschakki: Are you sure that the receivingchain ID is not required on the sidechain? In the readme of the interop example, it looks like both config have the value: https://github.com/LiskHQ/lisk-sdk/blob/development/examples/interop/README.md

docs/modules/ROOT/pages/run-blockchain/setup-relayer.adoc
TalhaMaliktz: This plugin is registered in the `app.ts` as per the following guide:
https://lightcurve.atlassian.net/wiki/spaces/SDK/pages/6095082/Chain+Connector+Plugin+Usage?NO_SSR=1

docs/modules/ROOT/pages/run-blockchain/setup-relayer.adoc
TalhaMaliktz: AFAIK there is no CCM after registration. @ishantiw maybe can help us clarify this?

docs/modules/ROOT/pages/run-blockchain/setup-relayer.adoc
TalhaMaliktz: Yes, Tried the example myself. Didn't throw any errors. For the mainchain, it does throw an error if you don't have a chainID in the config. The same is mentioned in the following guide as well.
https://lightcurve.atlassian.net/wiki/spaces/SDK/pages/6095082/Chain+Connector+Plugin+Usage?NO_SSR=1

Tschakki: Can you please point me to where it is mentioned exactly? And why does the guide say `chainID`, while the Readme says `receivingChainID`?

TalhaMaliktz: In this Prepare config section (the guide is focused on a sidechain), and yes it should be `receivingChainID` 👍🏼 
https://lightcurve.atlassian.net/wiki/spaces/SDK/pages/6095082/Chain+Connector+Plugin+Usage?NO_SSR=1#Prepare-config-for-chain-connector-plugin

docs/modules/ROOT/pages/run-blockchain/setup-relayer.adoc
Tschakki: I am pretty sure there is: https://liskhq.github.io/lisk-docs/beta/understand-blockchain/interoperability/sidechain-registration-and-recovery.html#registration-ccm

docs/modules/ROOT/pages/run-blockchain/setup-relayer.adoc
Tschakki: I think the guide is not  following best practise in this case :)

docs/modules/ROOT/pages/run-blockchain/setup-relayer.adoc
TalhaMaliktz: Is there an endpoint or method which can be used to get details of such CCM?

docs/modules/ROOT/pages/run-blockchain/setup-relayer.adoc
TalhaMaliktz: Let me try registering it in the `plugin.ts` if it works then I can make the suggested change.

docs/modules/ROOT/pages/run-blockchain/setup-relayer.adoc
TalhaMaliktz: Please share the links that you think are suitable here:
> Managing a node connected to the mainchain (link to Lisk Core)
Managing a node connected to the sidechain

docs/modules/ROOT/pages/run-blockchain/setup-relayer.adoc
Tschakki: You can check the outbox size of the channel data store: https://liskhq.github.io/lisk-docs/beta/build-blockchain/register-sidechain.html#check-the-sidechain-channel 

Not sure if you can look at the individual CCMs, though 🤔

docs/modules/ROOT/pages/run-blockchain/setup-relayer.adoc
Tschakki: 1. I would link to this section: run-blockchain/index.html#how-to-set-up-a-lisk-mainnet-node
2. I am not sure we have a fitting guide for this. Also, the setup might be different on different sidechains, so maybe we can just add a footnote to refer to the setup guides of the respective sidechain.

docs/modules/ROOT/pages/run-blockchain/setup-relayer.adoc
TalhaMaliktz: Hmmm! I do get the size of the outbox: 1, not sure what it means for us though.

**Sidechain**
```
./bin/run endpoint:invoke interoperability_getChannel '{"chainID": "04000000"}' --pretty                                                                              
{
  "messageFeeTokenID": "0400000000000000",
  "outbox": {
    "appendPath": [
      "b947f78350104a72381044ae1a239752084215e0d5e2fc1cb0eba8956ec4761a"
    ],
    "root": "b947f78350104a72381044ae1a239752084215e0d5e2fc1cb0eba8956ec4761a",
    "size": 1
  },
  "inbox": {
    "appendPath": [],
    "root": "e3b0c44298fc1c149afbf4c8996fb92427ae41e4649b934ca495991b7852b855",
    "size": 0
  },
  "partnerChainOutboxRoot": "e3b0c44298fc1c149afbf4c8996fb92427ae41e4649b934ca495991b7852b855",
  "minReturnFeePerByte": "1000"
}
```
**Mainchain**
```
lisk-core endpoint:invoke interoperability_getChannel '{"chainID": "04000002"}' --pretty                                                                               
{
  "messageFeeTokenID": "0400000000000000",
  "outbox": {
    "appendPath": [
      "1a8fd0bea33f8144853e4f86ea57c38e84046dd7def2776e6d6782b90357eae8"
    ],
    "root": "1a8fd0bea33f8144853e4f86ea57c38e84046dd7def2776e6d6782b90357eae8",
    "size": 1
  },
  "inbox": {
    "appendPath": [],
    "root": "e3b0c44298fc1c149afbf4c8996fb92427ae41e4649b934ca495991b7852b855",
    "size": 0
  },
  "partnerChainOutboxRoot": "e3b0c44298fc1c149afbf4c8996fb92427ae41e4649b934ca495991b7852b855",
  "minReturnFeePerByte": "1000"
}
```

There is no success message for me, I get the error message. ⬇️
> In case the default scenario is, that there are CCMs to be processed, so that it doesn't log the described error, it should explain the success message instead.

docs/modules/ROOT/pages/run-blockchain/setup-relayer.adoc
TalhaMaliktz: I have made the adjustment to the text.

docs/modules/ROOT/pages/run-blockchain/setup-relayer.adoc
Tschakki: @ishantiw please clarify if this is expected behaviour of the chain connector plugin, and how to verify that the plugin was enabled successfully. Imho, the user should not wait for an error to show up, because this is not a sign that something works, but more that something doesn't work as expected. So there should be some additional log messages concerning a successfully sent CCU or some endpoints to query the required information to verify the plugin is working as expected.

docs/modules/ROOT/pages/run-blockchain/setup-relayer.adoc
```diff
@@ -0,0 +1,352 @@
...
+--
+Sample configuration::
++
+--
+[source,json]
+----
+{
+    // [...]
+    "plugins": {
+        "chainConnector": {
```
ishantiw, line 143: maybe mention `receivingChainID` as its mandatory field

docs/modules/ROOT/pages/run-blockchain/setup-relayer.adoc
```diff
@@ -0,0 +1,352 @@
...
+--
+[source,bash]
+----
+lisk-core console
+Entering Lisk REPL: type `Ctrl+C` or `.exit` to exit
+----
+--
+=====
+
+. Call the `encrypt` function to create an encrypted private key and pass the `generatorPrivateKey` of your account to it.
```
ishantiw:
```suggestion
. Call the `encrypt` function to create an encrypted private key and password the `generatorPrivateKey` of your account to it.
```

ishantiw: Ah this is not generatorPrivateKey but ed private key, you can simply say to encrypt private key of any account on the other chain that you want to use as a relayer address and has balance to sign and send the transaction

docs/modules/ROOT/pages/run-blockchain/setup-relayer.adoc
```diff
@@ -0,0 +1,352 @@
...
+----
+--
+=====
+
+
+. The `encryptMessageWithPassword` function will return an encrypted key, which should be added to the config of the blockchain.
+
+=== Prepare config for chain connector plugin
+There's a slight difference in the `chainConnector` config of a sidechain and a mainchain. 
+The mainchain expects the `chainID` of the receiving chain along with the mandatory fields.
```
ishantiw: on both mainchain and sidechain you need chainID of the receiving chain

docs/modules/ROOT/pages/run-blockchain/setup-relayer.adoc
```diff
@@ -0,0 +1,352 @@
...
+[tabs]
+=====
+Sidechain::
++
+--
+On a sidechain, the plugin can be enabled using the `--enable-chain-connector-plugin` flag whilst starting the sidechain client.
+
+[source,bash]
+----
+./bin/run --enable-chain-connector-plugin --overwrite-config
```
ishantiw:
```suggestion
./bin/run start --enable-chain-connector-plugin --overwrite-config
```

docs/modules/ROOT/pages/run-blockchain/setup-relayer.adoc
```diff
@@ -0,0 +1,352 @@
...
+
+[source,bash]
+----
+npm run build
+----
+The client can then be started with the following command:
+
+[source,bash]
+----
+./bin/run --overwrite-config
```
ishantiw:
```suggestion
./bin/run start --overwrite-config
```

docs/modules/ROOT/pages/run-blockchain/setup-relayer.adoc
ishantiw: When we register a chain, a registration CCM is created and added to the outbox but if there is no valid certificate to be created, then we wait for for the finalized height, meanwhile log "No valid CCU can be generated for the height: <>". Once the first finalized height is created then we can include the CCM with the first CCU and send it across, for subsequent CCUs, we can have a scenario where there are pending CCMs but no certificate, in that case we can still send a CCU with only pending CCMs and without certificate.

ishantiw (changes requested):

docs/modules/ROOT/pages/run-blockchain/setup-relayer.adoc
Tschakki: Sorry, I do not see the requirement to have a running node in the text.
Also, I don't see why the sidechain relayer has prerequisite to habe the status `active` and `registered`, but not the mainchain relayer. Imho there is no need to separate between the two. For both, it is necessary to complete the registration process as explained in the guide (which already includes both, sc and mc registration)

docs/modules/ROOT/pages/run-blockchain/setup-relayer.adoc
```diff
@@ -0,0 +1,359 @@
...
+
+The chain connector plugin comes pre-installed with any chain which has been initialized with Lisk SDK v6.
+However, in case it is required to install it manually, run the following command in the root folder of the client.
+
+[source,bash]
+----
+npm install @liskhq/lisk-framework-chain-connector-plugin
+----
+
+== Configuring the Chain Connector plugin
```
Tschakki, line 53: This is more like an explanation, and there are no concrete steps that a user can do here. I would therefore merge this content with the section where we are actually configuring the plugin, under `=== Prepare config for chain connector plugin`

docs/modules/ROOT/pages/run-blockchain/setup-relayer.adoc
```diff
@@ -0,0 +1,359 @@
...
+====
+To use this guide, it is assumed that the following criteria have been met:
+
+.For a sidechain to set up a relayer node:
+* The sidechain must be registered to the mainchain as described in the xref:{url_guides_reg_sidechain}[] guide.
+* The sidechain account on the mainchain must have either the `registered` or `active` status.
+
+.For a mainchain to set up a relayer node:
+* The mainchain should be registered on the sidechain to which it intends to communicate as described in the xref:{url_guides_reg_mainchain}[How to register the mainchain on the sidechain] guide{fn_mainnet}.
+
```
Tschakki: Sorry, I do not see the requirement to have a running node in the text.
Also, I don't see why the sidechain relayer has prerequisite to have the status `active` and `registered`, but not the mainchain relayer. Imho there is no need to separate between the two. For both, it is necessary to complete the registration process as explained in the guide (which already includes both, sc and mc registration)

docs/modules/ROOT/pages/run-blockchain/setup-relayer.adoc
Tschakki: Just to clarify, so in case the sidechain is newly registered, and we set up the first relayer, then this log message is expected, right? But for a running sidechain, a new relayer would be already able to send CCUs because the finalized height is reached? I think we need to find and explain a way to verify that the relayer was set up correctly for every use case. This guide is not only for people setting up a relayer after registration, but more how to setup a relayer in general.

Tschakki (changes requested):

docs/modules/ROOT/pages/run-blockchain/setup-relayer.adoc
TalhaMaliktz: > Sorry, I do not see the requirement to have a running node in the text.

Without a running node, it will not be possible to even do the registration. The reader is already aware or should be aware of running a node, before setting up the relayer node. IMHO this is a pre-req for the registration guide.

> Also, I don't see why the sidechain relayer has a prerequisite to have the status active and registered, but not the mainchain relayer. Imho there is no need to separate the two. 

Because the liveness condition is only for a sidechain and not for a mainchain. A mainchain is going to stay alive.
https://github.com/LiskHQ/lips/blob/main/proposals/lip-0045.md#life-cycle-of-a-sidechain

The separation of prerequisites is intentional as either a sidechain node operator will read it or a mainchain node operator, to set up the relayer. That's why content is separated for mainchain and sidechain, throughout the page.

docs/modules/ROOT/pages/run-blockchain/setup-relayer.adoc
TalhaMaliktz: I tried that and this breaks the flow for me as in the suggested case, the introduction of the concept `encryptedPrivateKey` will come after the instruction: **Creating an encrypted private key**. If you agree, I would like to keep it as is.

docs/modules/ROOT/pages/run-blockchain/setup-relayer.adoc
TalhaMaliktz: @Tschakki Good point. I agree. @ishantiw Is there a way to realize what Mona suggests?

Also if you guys agree, I think we should have a separate guide about sending CCU/CCMs and should also describe any edge cases a node operator has to face. This guide can be part two of the set-up relayer guide.

Tschakki: > separate  guide about sending CCU/CCMs

I don't think there are many manual steps included after activating the plugin - what exactly do you imagine to be explained in the guide?

docs/modules/ROOT/pages/run-blockchain/setup-relayer.adoc
Tschakki: > Without a running node, it will not be possible to even do the registration. The reader is already aware or should be aware of running a node, before setting up the relayer node. IMHO this is a pre-req for the registration guide.

Just to be clear here: Completing the registration guide is NOT a prerequisite for a relayer, to setup a relayer. It is enough if SOMEONE has registered the sidechain before. So it cannot be assumed that every relayer went through the registration guide. So this is why I would make the changes to the prerequisites as I suggested.

And about making different prerequisites here for mainchain and sidechain, I do not see the benefit of separating them, because no one will only register mainchain to sidechain, without doing the second step. Both registration belong together. So for both relayer on mainchain and sidechain, it should be prerequisite to finish BOTH registrations.
This is why i would only refer to the registration guide as a whole as a prerequisite.

docs/modules/ROOT/pages/run-blockchain/setup-relayer.adoc
TalhaMaliktz: Not sure, as I am not familiar with the process. Discussing various ways of sending CCU/CCM (if there are any) can be explained there. Edge cases can be addressed as well. We can also show various responses like for a successful CCU/CCM, for failure, etc. More clarity can be achieved once we work on the topic.

docs/modules/ROOT/pages/run-blockchain/setup-relayer.adoc
```diff
@@ -0,0 +1,359 @@
...
+[NOTE]
+====
+To use this guide, it is assumed that the following criteria have been met:
+
+.For a sidechain to set up a relayer node:
+* The sidechain must be registered to the mainchain as described in the xref:{url_guides_reg_sidechain}[] guide.
+* The sidechain account on the mainchain must have either the `registered` or `active` status.
+
+.For a mainchain to set up a relayer node:
+* The mainchain should be registered on the sidechain to which it intends to communicate as described in the xref:{url_guides_reg_mainchain}[How to register the mainchain on the sidechain] guide{fn_mainnet}.
```
TalhaMaliktz: @Tschakki  Does this sound okay to you?
```suggestion
.Prerequisites
[NOTE]
====
To use this guide, it is assumed that the following criteria have been met:

* The sidechain must be registered to the mainchain as described in the xref:{url_guides_reg_sidechain}[] guide.
* The sidechain account on the mainchain{fn_mainnet} must have either the `registered` or `active` status.
```

docs/modules/ROOT/pages/run-blockchain/setup-relayer.adoc
TalhaMaliktz: I have made a suggestion above, please consider that.

docs/modules/ROOT/pages/run-blockchain/setup-relayer.adoc
Tschakki: It is going in the right direction, but as mentioned before, I would put in the prerequisites that they need to have nodes on sidechain and mainchain.

1. Managing a node connected to the mainchain (link to Lisk Core)
2. Managing a node connected to the sidechain
3. The sidechain should be registered to the mainchain according to the guide "How to register a sidechain"(link) and the status of the sidechain account should be `registered` or `active`.

TalhaMaliktz: Done.

docs/modules/ROOT/pages/run-blockchain/setup-relayer.adoc
```diff
@@ -0,0 +1,357 @@
...
+--
+[source,bash]
+----
+2023-03-17T14:42:30.426Z INFO XYZ.local plugin_chainConnector 96899 No valid CCU can be generated for the height: 58
+----
+--
+=====
+Since we just set up a relayer node and haven't sent a CCU/CCM, the aforementioned log messages are expected.
+
+The messages suggest that the blockchain doesn't have any finalized block height for which we can create a certificate and there are no pending CCUs/CCMs to send across the chain.
```
Tschakki: Added a sentence to explain next step after the error message. Maybe it can be enhanced with a first log message of a successfully sent CCU.

```suggestion
The messages suggest that the blockchain doesn't have any finalized block height for which we can create a certificate and there are no pending CCUs/CCMs to send across the chain.

The relayer node will start relaying CCUs to the receiving chain, once the finalized height is reached.
```

docs/modules/ROOT/pages/run-blockchain/setup-relayer.adoc
Tschakki: thanks 👍

docs/modules/ROOT/pages/run-blockchain/setup-relayer.adoc
Tschakki: Yes I agree, let's keep that in mind and ocme back later to the topic, once we have more practical knowledge :)

Tschakki (approved):
Looks great! :) Just a few minor comments.

docs/modules/ROOT/pages/run-blockchain/setup-relayer.adoc
```diff
@@ -0,0 +1,359 @@
...
+However, in case it is required to install it manually, run the following command in the root folder of the client.
+
+[source,bash]
+----
+npm install @liskhq/lisk-framework-chain-connector-plugin
+----
+
+== Configuring the Chain Connector plugin
+Once the plugin is installed, various configurations are available for the node operator to set.
+With these configurations, the details such as the receiver’s chainID, the receiving chain’s path, CCU frequency, encrypted private key, etc. can be set.
```
ishantiw:
```suggestion
With these configurations, the details such as the receiving chain’s chainID and it's IPC path or WS URL, CCU frequency, encrypted private key, etc. can be set.
```

docs/modules/ROOT/pages/run-blockchain/setup-relayer.adoc
```diff
@@ -0,0 +1,359 @@
...
+|string
+|The fee to be paid for each CCU transaction
+|100000000
+
+|`password`
+|string
+|Password to decrypt encryptedPrivateKey
+|lisk
+
+|`isSaveCCU`
```
ishantiw, line 109: I have recently added ccuSaveLimit which by default is 300 CCUs and user can set it to any value to tell how many CCUs it wants to save https://github.com/LiskHQ/lisk-sdk/pull/8315

docs/modules/ROOT/pages/run-blockchain/setup-relayer.adoc
```diff
@@ -0,0 +1,359 @@
...
+--
+[source,bash]
+----
+2023-03-17T14:42:30.426Z INFO XYZ.local plugin_chainConnector 96899 No valid CCU can be generated for the height: 58
+----
+--
+=====
+Since we just set up a relayer node and haven't sent a CCU/CCM, the aforementioned log messages are expected.
+
+The messages suggest that the blockchain doesn't have any finalized block height for which we can create a certificate and there are no pending CCUs/CCMs to send across the chain.
```
ishantiw:
```suggestion
The messages suggest that the blockchain doesn't have any finalized block height for which we can create a certificate or there are no pending CCUs/CCMs to send across the chain.
```

ishantiw (commented):
Looks good, just a few more minor comments

ishantiw (approved):

PR #1752, opened 2023-03-22
Update decoding encoding guide

### LINKED ISSUE #1581: Update page `Decoding & encoding` data
### Description

- Update to use Lisk SDK v6

Tschakki:
Closes #1581


shuse2 (approved):

TalhaMaliktz (changes requested):
The existing content looks great. However, I think we should describe the usage of the `@liskhq/lisk-codec` package as well. 

Especially the cases where we encode and decode data using custom schemas. For example, decoding the data from blockchain events, etc., and any other use case for encoding data as well.

The example of decoding is already available [here](https://liskhq.github.io/lisk-docs/beta/build-blockchain/plugin/plugin-class.html#sync-and-store-new-event)
We will need an example for encoding data using the `codec`.

Tschakki:
@TalhaMaliktz yes I see your point, I agree it would be a good addition to the guide. Would you mind if we create a new issue for this, to tackle this in another PR? As this requires quite some new content, I think it would go beyond the scope of this issue, which was mainly about updating the existing content.

TalhaMaliktz (approved):
@Tschakki Yes, creating a new issue sounds like a plan. I am approving the PR 🎉

Tschakki:
Thank you, I have created the issue here: https://github.com/LiskHQ/lisk-docs/issues/1756

PR #1785, opened 2023-05-15
Add glossary tooltips

### LINKED ISSUE #1321: Add glossary tooltips
### Description
Add automatic tooltips for glossary terms.

### Motivation
Having glossary term automatically linked with tool tips would decrease the effort of maintaining the links, and also it would no break the readers flow, because they can view the explanation of the term on the same page.

### Additional information

Try: https://gitlab.com/djencks/asciidoctor-glossary

Tschakki:
This is a very simple implementation for tooltips into the UI. Currently, I think it is mainly suitable for abbreviations and small remarks, maybe it can be improved/updated in the future.
For the classical glossary entries, I am not sure if we still need a separate glossary, or if it is possible to explain all important terms inside of the documentation, which I think should be the preferred way to explain terminology.

In the current implementation, the tooltips are generated from a file `glossary.json`, and content on every page is filtered for the terms. A tooltip is added for each first occasion of the term on each page, excluding headings and the ToC.

Closes #1321


ui/src/css/extensions/custom.css
```diff
@@ -497,6 +497,53 @@
...
+  position: absolute;
+  top: 100%;
+  left: 50%;
+  margin-left: -5px;
+  border-width: 5px;
+  border-style: solid;
+  border-color: #555 transparent transparent transparent;
+}
+
+/* Show the tooltip text when you mouse over the tooltip container */
```
chris529:
```suggestion
/* Show the tooltip text when you hover your mouse over the tooltip container */
```

TalhaMaliktz (changes requested):
- Good addition to the documentation definitely improves the visibility 🎉 
- For glossary: Maybe in the future, we can try to imitate an on-hover pop-up containing the definition for the word, as we see in Wikipedia.

image

- Also, the tooltip filtration needs tweaking as it highlights partial words as well.

image

chris529 (commented):
Looks great! Only left one comment..👍

TalhaMaliktz (approved):
LGTM

chris529 (approved):
LGTM! 👍

PR #1848, opened 2023-07-26
Mention using a different port when starting the app with the macOS in the nft tutorial in lisk docs

### LINKED ISSUE #1499: Mention to use different port when starting the app on macOS
### Description

In the NFT tutorial, it should be noted, that the application needs to start with a different port, because 5000 is already in use in latest macOS. This needs to be updated in the current docs version.

### Affected Pages

`tutorial/nft.adocl#registering-module-and-plugin`

### Affected Branch 

`main`

chris529:
Closes #1499

### DIFF
--- a/docs/modules/ROOT/pages/tutorial/nft.adoc
+++ b/docs/modules/ROOT/pages/tutorial/nft.adoc
@@ -1403,7 +1403,7 @@ Save and close the `index.js`.
 
 Now when the application is started again with `node index.js`, the blockchain application will load the newly created NFT module and the plugins, and the new features will become available to the blockchain application.
 
-NOTE: When using the macOS, please ensure a different port, for example, such as 8756 or 4724, as port 5000 is already in use with the latest macOS from v12.xx Monterey and above.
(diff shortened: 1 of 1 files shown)


docs/modules/ROOT/pages/tutorial/nft.adoc
```diff
@@ -1403,6 +1403,8 @@ Save and close the `index.js`.
 
 Now when the application is started again with `node index.js`, the blockchain application will load the newly created NFT module and the plugins, and the new features will become available to the blockchain application.
 
+NOTE: When using the macOS, please ensure a different port is used, for example, such as 8756 or 4724, as port 5000 is already in use with the latest macOS from v12 Monterey and above.
```
TalhaMaliktz:
```suggestion
NOTE: When using the macOS, please ensure that a different port other than `5000` is used, for example, `8756` or `4724`, as port `5000` is already used by the latest macOS starting from v12 Monterey and above.
```

TalhaMaliktz (approved):
Thankyou!

priojeetpriyom (approved):

PR #1864, opened 2023-08-11
Align code snippets

### LINKED ISSUE #1828: Optimize command examples (SDK)
The description of the issue is described in the [epic](https://github.com/LiskHQ/lisk-docs/issues/1490).

Tschakki:
Closes #1828

### DIFF
--- a/modules/ROOT/pages/client-cli.adoc
+++ b/modules/ROOT/pages/client-cli.adoc
@@ -639,6 +639,7 @@ EXAMPLES
 ./bin/run endpoint:invoke chain_getLastBlock --pretty
 ----
 
+.Response
 [source,json]
 ----
 {
--- a/modules/ROOT/pages/client-cli.adoc
+++ b/modules/ROOT/pages/client-cli.adoc
@@ -82,7 +82,11 @@ EXAMPLES
 ====
 [source,bash]
 ----
-$ ./bin/run autocomplete bash
+./bin/run autocomplete bash
+----
+
+.Response
+----
 Building the autocomplete cache... done
 
 Setup Instructions for DPOS-MAINCHAIN CLI Autocomplete ---
@@ -126,7 +130,10 @@ EXAMPLES
 ====
 [source,bash]
 ----
-$  ./bin/run console                             
+./bin/run console
+----
+
+----
 Entering Lisk REPL: type `Ctrl+C` or `.exit` to exit
 hello-client >
 ----
@@ -159,7 +166,12 @@ EXAMPLES
 ====
 [source,bash]
 ----
-$./bin/run hash-onion --count=10 --distance=2 --pretty 
+./bin/run hash-onion --count=10 --distance=2 --pretty
+----
+
+.Response
+[source,json]
+----
 {
   "count": 10,
   "distance": 2,
@@ -228,7 +240,7 @@ EXAMPLES
 ====
 [source,bash]
 ----
-$ ./bin/run start
+./bin/run start
 ----
 ====
 
@@ -274,7 +286,12 @@ EXAMPLES
 ====
 [source,bash]
 ----
-$ ./bin/run block:get 2 --pretty
+./bin/run block:get 2 --pretty
+----
+
+.Response
+[source,json]
+----
 {
   "header": {
     "version": 2,
@@ -350,11 +367,7 @@ EXAMPLES
 ====
 [source,bash]
 ----
-$ ./bin/run blockchain:export       
-Exporting blockchain:
-   /Users/xyz/.lisk/dpos-mainchain/data/blockchain.db
-Export completed:
-   /Users/xyz/Documents/GitHub/lisk-sdk/examples/dpos-mainchain/blockchain.db.tar.gz
+./bin/run blockchain:export
 ----
 ====
 
@@ -382,8 +395,7 @@ EXAMPLES
 ====
 [source,bash]
 ----
-$ ./bin/run blockchain:hash --data-path /Users/xyz/Documents/GitHub/lisk-sdk/examples/dpos-mainchain/blockchain.db      
-e3b0c44298fc1c149afbf4c8996fb92427ae41e4649b934ca495991b7852b855
+./bin/run blockchain:hash --data-path /Users/xyz/Documents/GitHub/lisk-sdk/examples/dpos-mainchain/blockchain.db
 ----
 ====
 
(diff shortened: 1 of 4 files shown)


modules/ROOT/pages/client-cli.adoc
```diff
@@ -82,7 +82,11 @@ EXAMPLES
 ====
 [source,bash]
 ----
-$ ./bin/run autocomplete bash
+./bin/run autocomplete bash
+----
+
+.Response
+----
```
TalhaMaliktz, line 89: Please use `[source,bash]` for the responses as well, since the output will also be visible in the bash. Secondly, to stay in sync with what we have done for `docs-service`. I suggest using captions with each snippet to improve the outlook.

PS: This applies to the entire page

modules/ROOT/pages/client-cli.adoc
```diff
@@ -126,7 +130,10 @@ EXAMPLES
 ====
 [source,bash]
 ----
-$  ./bin/run console                             
+./bin/run console
+----
+
+----
```
TalhaMaliktz, line 136: Same comment as above!

modules/ROOT/pages/client-cli.adoc
```diff
@@ -626,7 +636,11 @@ EXAMPLES
 ====
 [source,bash]
 ----
-$./bin/run endpoint:invoke chain_getLastBlock --pretty
+./bin/run endpoint:invoke chain_getLastBlock --pretty
+----
+
```
TalhaMaliktz, line 641:
```suggestion

.Response
```

modules/ROOT/pages/client-cli.adoc
```diff
@@ -710,8 +724,14 @@ EXAMPLES
 [source,bash]
 ----
 ./bin/run generator:disable lskqaxxmj78frvgpjgwvf4yqjjkcrr9yhn2sxxwm3
+----
+
+----
 ? Enter password to decrypt the encrypted passphrase:  ********
+----
```
TalhaMaliktz, line 732: Same comment: Highlight as bash, and use captions to differentiate between lines.

modules/ROOT/pages/client-cli.adoc
```diff
@@ -710,8 +724,14 @@ EXAMPLES
...
 
+.Response
+----
```
TalhaMaliktz, line 735: Same comment as above

modules/ROOT/pages/client-cli.adoc
```diff
@@ -754,12 +774,16 @@ EXAMPLES
 [source,bash]
 ----
 ./bin/run generator:enable lskqaxxmj78frvgpjgwvf4yqjjkcrr9yhn2sxxwm3 --use-status-value
-
+----
+----
```
TalhaMaliktz, line 779: Would be nice to have caption here describing, what's happening below.

modules/ROOT/pages/references/lisk-commander/index.adoc
```diff
@@ -53,6 +53,10 @@ Confirm that Node.js and NPM have been successfully installed by executing the f
 [source,bash]
```
TalhaMaliktz, line 54: **LINE 51:**
Confirm that Node.js and NPM have been successfully installed by executing the following commands:

modules/ROOT/pages/references/lisk-commander/sensitive-inputs.adoc
```diff
@@ -17,7 +17,11 @@ Please note that this approach is only available for passphrases, and not other
 
 [source,bash]
 ----
-$ lisk message:encrypt bba7e2e6a4639c431b68e31115a71ffefcb4e025a4d1656405dfdcd8384719e0 'Hello world'
+lisk message:encrypt bba7e2e6a4639c431b68e31115a71ffefcb4e025a4d1656405dfdcd8384719e0 'Hello world'
+----
+
```
TalhaMaliktz, line 22: A caption here would help with the outlook.

TalhaMaliktz (changes requested):

chris529 (commented):
I concur with Talha's suggestions, otherwise all looks good 👍

modules/ROOT/pages/client-cli.adoc
Tschakki: Specifying bash as source highlighter only makes sense if there are bash commands that should be highlighted correctly. If plain text is returned in the command line, i don't see the benefit to mark it as bash, as it might highlight words that are not supposed to be highlighted in the plain text.

Tschakki: Concerning the captions, i am not sure what you mean? I added a lot of captions already

modules/ROOT/pages/client-cli.adoc
Tschakki: Imho it is self-explanatory what is happening.

And it is just a reference, I wouldn't use too much words and explanations here. but if you have a short and descriptive caption in mind please share, then i can add it :)

modules/ROOT/pages/references/lisk-commander/sensitive-inputs.adoc
Tschakki: What caption do you have in mind?

modules/ROOT/pages/client-cli.adoc
TalhaMaliktz: I see your point! Considering your thoughts, I suggest we make a rule of thumb here! Use `bash` for **Commands** and `bash` or `JSON` for **Responses** and can have the simple text highlighter for cases where a user has to input some info into the terminal. For example, The following style guide:

image

There are places where the response formatting doesn't match with the other (Example Line # 823 & 969), so I think, we can use a consistent convention throughout the page and the docs. If we agree then, I can replicate this style guide for optimizing snippets from `lisk-docs` as well, so that everything is in sync.

What do you say?

modules/ROOT/pages/references/lisk-commander/sensitive-inputs.adoc
TalhaMaliktz: I think the suggestion above regarding formatting could mitigate the need for a caption here IMHO! The only reason I suggest the addition of a caption here is that there are two snippets with the same styling and to me, something seems missing aesthetically. Different formatting can also serve the purpose.

modules/ROOT/pages/client-cli.adoc
Tschakki: Ah ok, yes when it is for the styling, I think we can do that, bash doesn't really highlight much anyway.

Another option we could consider is to update the docs UI so that plain text code snippets have the black background as well. What do you think?

TalhaMaliktz: Yeah! That could work as well. Will we be using a specific keyword in the markdown to indicate the use of this particular style?

TalhaMaliktz (approved):
LGTM

Issue #1924, opened 2023-12-05
Fix formula for module store keys


Tschakki:
### Description
As per LIP 40, key for module stores is a combination of `storePrefix` + `substorePrefix` + `sha256(storeKey)`,

the formula on the StateMachine page should be updated accordingly.

### Affected branch / URL
https://lisk.com/documentation/beta/understand-blockchain/state-machine.html#the-blockchain-state-as-sparse-merkle-tree

PR #1934, opened 2023-12-13
Fix formula for module store keys in Lisk docs

chris529:
#1924

### DIFF
--- a/docs/modules/ROOT/pages/understand-blockchain/state-machine.adoc
+++ b/docs/modules/ROOT/pages/understand-blockchain/state-machine.adoc
@@ -68,13 +68,13 @@ Each module registered on the chain defines its own generic key-value map, where
 * *key*: Store Prefix + Substore prefix + sha256(Store key)
 
 ** **Store prefix**: The module store prefix is calculated automatically from the first 4 bytes of the hash of the module name, where the first bit of the store prefix is set to 0.
-The Module name within the Lisk framework is the unique identifier of the module, identifying the key-value map.
+The module name within the Lisk framework is the unique identifier of the module, identifying the key-value map.
 It is a unique string that is used to construct store prefix and various module-related operations.
 On the other hand, the Interoperability module store prefix is set to a constant, with the first bit set to 1.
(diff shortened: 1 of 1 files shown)


TalhaMaliktz (approved):
LGTM 👍🏼

docs/modules/ROOT/pages/understand-blockchain/state-machine.adoc
```diff
@@ -61,17 +62,19 @@ A blockchain created with the Lisk SDK defines a single global store.
...
+
 Each module registered on the chain defines its own generic key-value map, where:
 
-* *key*: Module Name + Store prefix + Substore prefix + Store key
-** **Module Name**: The unique identifier of the module, identifying the key-value map.
-The module name is a unique string that is used to construct store prefix and various module-related operations.
+* *key*: Store Prefix + Substore prefix + sha256(Store key)
+
 ** **Store prefix**: The module store prefix is calculated automatically from the first 4 bytes of the hash of the module name, where the first bit of the store prefix is set to 0.
+The module name within the Lisk framework is the unique identifier of the module, identifying the key-value map.
```
Incede:
```suggestion
The module name is the unique identifier of the module, identifying the key-value map.
```
Not a must but if you agree I'd omit `within the Lisk framework` as it seems a bit unnecessary and could be a bit confusing in terms of SDK vs framwework

docs/modules/ROOT/pages/understand-blockchain/state-machine.adoc
```diff
@@ -61,17 +62,19 @@ A blockchain created with the Lisk SDK defines a single global store.
...
+It is a unique string that is used to construct store prefix and various module-related operations.
```
Incede: I think module name is now not used in "various module-related operations" ?

Also the part about `used to construct store prefix` is already indicated in the previous section about "Store Prefix" and if you agree I'd omit it.
```suggestion
```

Incede (approved):
LGTM. I just have a minor question and 2 optional suggestions.

docs/modules/ROOT/pages/understand-blockchain/state-machine.adoc
chris529: @Incede Yes, good point. Thanks Rishi 👍

PR #1936, opened 2023-12-15
Copy to clipboard renders JSON incorrectly in Lisk docs

### LINKED ISSUE #1917: Copy to clipboard renders JSON incorrectly
### Description
Copying the command to creating a stake transaction by using the "copy to clipboard" function from the [relevant](https://lisk.com/documentation/beta/run-blockchain/staking.html#create-a-stake-transaction) lisk-docs page and executing it results in the following error
`SyntaxError: Unexpected end of JSON input`

Copying the same command manually and then executing it does not result in a error

### Affected branch / URL
[URL(s) of the page(s) affected by the bug.](https://lisk.com/documentation/beta/run-blockchain/staking.html#create-a-stake-transaction)
I think other places also have the same problem

chris529:
Closes #1917


Incede (approved):
As [commented](https://github.com/LiskHQ/lisk-docs/issues/1917#issuecomment-1866983212) I could not reproduce the issue so no changes shall be needed.

Tschakki (approved):
🎉

PR #1937, opened 2023-12-19
Update code background color

### LINKED ISSUE #1915: The font color of code blocks is extremely light and should be changed to something stronger.
### Motivation
Whilst viewing the docs in light mode, the code snippet's color scheme a very light, specially, the light grey color of text in code snippets. This makes it hard to read the code snippets. The theme should be changed to something which has a stronger color palette for the code text.

Tschakki:
Closes #1915 

Tested: https://liskhq.github.io/lisk-docs/


TalhaMaliktz (changes requested):
The color chosen increases the visibility for JS, TS, and Bash scripts, however, in the sources where no formatting is mentioned, the visibility and readability need an improvement. Please consider the following two examples.

image

image

Also, I would choose a color from the Lisk Color palette, maybe the blue color for snippets in the dark mode can be chosen here as well.

Tschakki:
Good catch, I updated the css rule so it shouldn't include plain text snippets

TalhaMaliktz (approved):
Thanks! LGTM 👍🏼

chris529 (approved):
LGTM! 👍

